PR #6, opened 2022-09-13
add min-browser

NextWork123:


ptr1337 (approved):
Seems good.
Would be really good, if you would also maintain as soon as updates are done.
Merged and will be pushed into the repo

PR #7, opened 2022-10-03
New package: awesome-luajit

ventureoo:
I know CachyOS doesn't have an awesomeWM setup ready right now, but I'd still like to have this PKGBUILD in the repositories for the future if it shows up.

Luajit is just faster than regular versions of Lua, and within the whole WM as awesome I think performance is important. So, unlike the AUR package, the build with the latest luajit is fixed here.


ptr1337 (approved):
LGTM

PR #9, opened 2022-10-28
Improvements for OBS

SoongVilda:
OBS in the Arch Linux repo mission features and dependencies don't lead you in the right way - you must study Arch wiki, and what's next should be installed to OBS. So I decided to fix that.


ptr1337 (approved):
Seems good to me! Please maintain them as soon updates are out, so we keep them up2date in the repo.
Much thanks for your work!

PR #10, opened 2022-10-28
pipewire patch

SoongVilda:
Added patch for a pipewire audio capture.


ptr1337 (approved):
Next time also updated please the pkgrel. :)

PR #15, opened 2022-11-15
Intel QuickSync (QSV) - tigerlake and newer

SoongVilda:
oneVPL Library and oneVPL GPU Runtime are successors for Intel Media SDK.

Intel® oneVPL GPU Runtime is a Runtime implementation of [oneVPL](https://github.com/oneapi-src/oneVPL/) API for Intel Gen GPUs. Runtime provides access to hardware-accelerated video decode, encode and filtering.

Supported video encoders: HEVC, AVC, MPEG-2, JPEG, VP9
Supported video decoders: HEVC, AVC, VP8, VP9, MPEG-2, VC1, JPEG, AV1
Supported video pre-processing filters: Color Conversion, Deinterlace, Denoise, Resize, Rotate, Composition

### DIFF
--- a/intel-packages/intel-makedepends-stage-final/onevpl-intel-gpu/PKGBUILD
+++ b/intel-packages/intel-makedepends-stage-final/onevpl-intel-gpu/PKGBUILD
@@ -1,0 +1,34 @@
+# Maintainer: Daniel Bermond <dbermond@archlinux.org>
+
+pkgname=onevpl-intel-gpu
+pkgver=22.5.4
+pkgrel=1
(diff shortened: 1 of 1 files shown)


ptr1337:
Is this package differently then the one in the AUR?
So far I have watched it is equal. As long packages are equal to the aur, we simply let them maintain there and they get automatically build as soon a upgrade is found! :)

SoongVilda:
Yes, it's the same package.

Okay, that's nice!

ptr1337:
Then I'll close here and at the package to the repo. 
Thanks!: )

PR #25, opened 2022-12-22
Add a readme

xeome:


ptr1337 (approved):
Seems good to me!

Maybe we could put a logo like it is here. Besides that its fine!

https://github.com/CachyOS/linux-cachyos#cachyos

PR #28, opened 2023-01-20
New package: protonfixes-ge-git

ventureoo:
Still WIP, I'm still not sure if it will use the system versions of winetricks and libmspack or if we should use the bundled versions from the repository, so testing is welcome. 

**Note:** this package is only needed for regular versions of proton/proton-experimental, proton-ge-custom already contains protonfixes.

### DIFF
--- a/protonfixes-ge-git/.SRCINFO
+++ b/protonfixes-ge-git/.SRCINFO
@@ -1,0 +1,19 @@
+pkgbase = protonfixes-ge-git
+	pkgdesc = Protonfixes fork by GloriousEggroll (git)
+	pkgver = 1.0.15.r546.gfc74674
+	pkgrel = 1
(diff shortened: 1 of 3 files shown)


ventureoo:
Sadly, this package probably does not make any sense. Proton simply does not import protonfixes in user_settings because it is outside the sandbox where Proton is running. That is, we need protonfixes to be in the directory with Proton, as is the case with Proton-GE-Custom, then it will be able to import it, which means there is no point in packaging it separately.
```

Proton: ************************************************
Proton: THERE IS AN ERROR IN YOUR user_settings.py FILE:
Proton: No module named 'protonfixes'
Proton: ************************************************
```

FYI: https://github.com/simons-public/protonfixes/issues/168

I'll close it until I find a good workaround.

PR #29, opened 2023-01-22
New package: nvlax-git

ventureoo:
Although upstream seems to be dead, nvlax still works fine for new driver versions and we don't have to wait for an nvidia-patch.

I'm really not sure if we really need hooks for auto-apply.

@ptr1337


ptr1337 (approved):
Thanks! I will add it to the repo.

Users can simply install it then if they need it.

PR #33, opened 2023-02-04
Proton-cachyos: Automate The Date Value with the use of API Github

NextWork123:
Hello,this function retrieves the SHA value of the latest commit in the "experimental_7.0" branch of the Proton repository from Github API. It then extracts the date of the commit and formats it as "7.0-YYYYMM". Finally, it replaces any hyphens in the date with dots and sets it as the value of the pkgver variable.

Best Regards,
NextWorks


proton-cachyos/PKGBUILD
```diff
@@ -5,8 +5,10 @@
 _optimize=${_optimize-y}
 
 pkgname=proton-cachyos
-_srctag=7.0-20230201
-_commit=
+_commit=$(curl -s https://api.github.com/repos/ValveSoftware/Proton/commits?sha=experimental_7.0 | grep '"sha"' | head -n 1 | cut -d '"' -f 4)
```
ventureoo, line 8: Maybe we should just do it in the pkgver() function?

loathingKernel:
You won't gain anything by doing this except going by the daily releases instead of valve's tagged weeklies. You are opening the package to breakages that are for internal testing and QA. The dated commits are valve's own releases for proton-experimental.

Not to mention that the experimental branch gets forcepushed regularly and the date of the latest commit in that branch might be before the previous one.

NextWork123:
Ok, np i close that pull.

PR #36, opened 2023-03-05
systemd-boot-manager: Prefer asynchronous discard for Btrfs instead of normal discard

ventureoo:
Synchronized discard is terrible


SM9CC (approved):
Looks good, thanks!

PR #66, opened 2023-12-20
julia 1.10

lseman:


lseman:
Updating libgit2 patch.

lseman:
It is not compiling on docker.

PR #67, opened 2023-12-21
Julia 1.10 + PGO

lseman:


lseman:
@ptr1337 Julia 1.10 was finally released.
I will recompile the package with the new version to get if nothing broke since the last RC.
We can then upstream it, or wait to see if ArchLinux will release a new patch.

lseman:
It is working.

PR #74, opened 2023-12-25
test(sqlite3): adding sqlite-pgo draft for discussion

lseman:
Starting a discussion on how could we provide sqlite3-pgo PKGBUILD.

A good benchmark for databases that can be used for PGOing is:
[ClickBench](https://github.com/ClickHouse/ClickBench)

see: https://github.com/zamazan4ik/awesome-pgo/blob/main/sqlite.md

The database for sqlite3 is 15GB, the default test randomly sample 2.2GB.
We could provide those 2.2GB, which compressed are as follows:

```
.rw-r--r--  291M seman 23 Dec 23:39   mydb.tar.xz
.rw-r--r--  334M seman 23 Dec 20:55   mydb.zst
```

But, what about the license of the database? Can we host it?

Any ideas?

@ptr1337 @vnepogodin

### DIFF
--- a/sqlite3/.SRCINFO
+++ b/sqlite3/.SRCINFO
@@ -93,7 +93,7 @@ build() {
   mv hits.csv ClickBench/sqlite/hits.csv
   cd ClickBench/sqlite
   LLVM_PROFILE_FILE=${PGO_PROFILE_DIR}/_%m_%p.profraw $SQLITE_DIR/sqlite3 mydb < create.sql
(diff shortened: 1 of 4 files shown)


lseman:
New version that considers downloading 15GB, untested in the full workflow mode (i.e., downloading the 15GB), but tested separated.
Please test the full pipeline in the server before deploying.

ptr1337:
After a lot of testing this does build now correctly. 
Thanks for your great work!

PR #103, opened 2024-01-02
LZ4 PGO

lseman:
With a clever little profile generator trick.

### DIFF
--- a/lz4/PKGBUILD
+++ b/lz4/PKGBUILD
@@ -23,7 +23,7 @@ _generate_and_process_data() {
   local _filename="data_${_data_size}_${_compression_level}.bin"
   echo "Data size: $_data_size KB"
   # Generate data
-  $_dataset_generator -g$_data_size > $_filename
+  $DATASET_GENERATOR -g$_data_size > $_filename
 
   echo "Compressing and decompressing data"
   echo $_filename
--- a/lz4/PKGBUILD
+++ b/lz4/PKGBUILD
@@ -1,0 +1,132 @@
+# Maintainer: Sébastien Luttringer
+# PGO build: Laio O. Seman <laio [at] ieee.org>
+
+pkgname=lz4
+epoch=1
+pkgver=1.9.4
+pkgrel=2
+pkgdesc='Extremely fast compression algorithm'
+arch=('x86_64')
+url='https://lz4.github.io/lz4/'
+license=('GPL2')
+makedepends=('git' 'cmake' 'ninja')
+checkdepends=('diffutils')
+depends=('glibc')
+source=("git+https://github.com/lz4/lz4.git#tag=v$pkgver"
+        'merged.profdata')
+sha256sums=('SKIP')
+
+# Function to generate data and compress-decompress it
+generate_and_process_data() {
+    echo "Generating data for $1 KB with compression level $2"
+    DATA_SIZE=$1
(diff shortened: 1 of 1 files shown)


lz4/PKGBUILD
```diff
@@ -0,0 +1,127 @@
...
+url='https://lz4.github.io/lz4/'
+license=('GPL2')
+makedepends=('git' 'cmake' 'ninja')
+checkdepends=('diffutils')
+depends=('glibc')
+source=("git+https://github.com/lz4/lz4.git#tag=v$pkgver")
+sha256sums=('SKIP')
+
+# Function to generate data and compress-decompress it
+generate_and_process_data() {
```
ventureoo, line 19: The function should be local.

lz4/PKGBUILD
```diff
@@ -0,0 +1,127 @@
...
+
+pkgname=lz4
+epoch=1
+pkgver=1.9.4
+pkgrel=2
+pkgdesc='Extremely fast compression algorithm'
+arch=('x86_64')
+url='https://lz4.github.io/lz4/'
+license=('GPL2')
+makedepends=('git' 'cmake' 'ninja')
```
ventureoo, line 12: Missing dependencies for clang/llvm

lz4/PKGBUILD
```diff
@@ -0,0 +1,127 @@
...
+checkdepends=('diffutils')
+depends=('glibc')
+source=("git+https://github.com/lz4/lz4.git#tag=v$pkgver")
+sha256sums=('SKIP')
+
+# Function to generate data and compress-decompress it
+generate_and_process_data() {
+    echo "Generating data for $1 KB with compression level $2"
+    DATA_SIZE=$1
```
ventureoo, line 21: Arguments vars should be locale

lz4/PKGBUILD
ventureoo: btw, same for sqlite

lz4/PKGBUILD
ventureoo: I meant to just add underscore before the function name like arch package guidelines says. It is not necessary to use the local keyword here.

lz4/PKGBUILD
```diff
@@ -0,0 +1,127 @@
...
+makedepends=('git' 'cmake' 'ninja', 'llvm' 'clang')
+checkdepends=('diffutils')
+depends=('glibc')
+source=("git+https://github.com/lz4/lz4.git#tag=v$pkgver")
+sha256sums=('SKIP')
+
+# Function to generate data and compress-decompress it
+local generate_and_process_data() {
+    echo "Generating data for $1 KB with compression level $2"
+    DATA_SIZE=$1
```
ventureoo, line 21:
```suggestion
    local data_size=$1
```

lseman: all variables are now private inside the functions, unless they are not (accessing global ones)

ventureoo: > 

specifying an underscore before a function name is just a convention for naming functions called only inside a script. Variables inside a function do not need to have an underscore in front of them, they should be declared as local variables to avoid clogging up the global environment.

lseman:
PR drooped.

ventureoo:
Why? My comments are stylistic remarks, not blocking feedback.

ventureoo (approved):

PR #106, opened 2024-01-03
Merge chwd and chwd-db into one package

ventureoo:
Depends on: https://github.com/CachyOS/chwd/pull/66

P. S. I'm not sure if we should bump new pkgver or just increase pkgrel for this.


vnepogodin (approved):
LGTM

PR #107, opened 2024-01-03
Add LTO for OpenJDK

ventureoo:
OpenJDK officially supports building with LTO from Java 9 and up (https://bugs.openjdk.org/browse/JDK-8292226), but it is disabled by default for all architectures except ARM, and unlike other packages, enabling it via `options=(lto)` causes build errors (See similar bug in Gentoo - https://bugs.gentoo.org/833098). Instead, LTO optimizations should be enabled via a special flag `--with-jvm-features=link-time-opt`. The LTO flag has been used for OpenJDK in Gentoo for a long time, so I think we can add it to ours as well.


ventureoo:
JDK 11 doesn't seem to be very buildable with LTO. So I'll have to drop it. Debian has already done the same thing (https://metadata.ftp-master.debian.org/changelogs//main/o/openjdk-11/openjdk-11_11.0.21+9-1~deb11u1_changelog, yes they build OpenJDK with LTO too).

PR #112, opened 2024-01-06
feat(python): python-312 with optimizations + bolt + automatic little/big endian detection trick

lseman:


python-312/PKGBUILD
```diff
@@ -0,0 +1,91 @@
...
+    # Ensure that we are using the system copy of various libraries (expat, zlib and libffi),
+    # rather than copies shipped in the tarball
+    rm -rf Modules/expat
+    rm -rf Modules/_decimal/libmpdec
+}
+
+build() {
+    cd "${srcdir}/Python-${pkgver}"
+
+    endianness=$(python -c "import sys; print('little' if sys.byteorder == 'little' else 'big')")
```
ventureoo, line 39: Why not just `print(sys.byteorder)`?

python-312/PKGBUILD
```diff
@@ -0,0 +1,91 @@
...
+
+    # Set the ax_cv_c_float_words_bigendian variable based on the Python output
+    if [ "$endianness" == "little" ]; then
+        export ax_cv_c_float_words_bigendian=no
+    else
+        export ax_cv_c_float_words_bigendian=yes
+    fi
```
ptr1337, line 46: Well, if this detection is at buildtime (in our case on zen4 machine) people with big little wont benefit from it

lseman: Python 3.12 refuses to compile without it being explicity.

python-312/PKGBUILD
lseman: Updated.

python-312/PKGBUILD
```diff
@@ -0,0 +1,91 @@
...
+    # Ensure that we are using the system copy of various libraries (expat, zlib and libffi),
+    # rather than copies shipped in the tarball
+    rm -rf Modules/expat
+    rm -rf Modules/_decimal/libmpdec
+}
+
+build() {
+    cd "${srcdir}/Python-${pkgver}"
+
+    endianness=$(python -c "import sys; print(sys.byteorder)")
```
ventureoo, line 39:
```suggestion
    local endianness=$(echo -n I | od -to2 | head -n1 | cut -f2 -d " " | cut -c6) 
    if [ "$endian" == "1" ]; then
        export ax_cv_c_float_words_bigendian=no
    else
        export ax_cv_c_float_words_bigendian=yes
    fi
```

This way will allow us to get rid of python in `makedepends` (which is missing btw).

PR #120, opened 2024-01-14
change(python): flags for perf

lseman:
New Python 3.12 has improved supported for perf, for that to works as expected we need to add -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer.


ptr1337:
> New Python 3.12 has improved supported for perf, for that to works as expected we need to add -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer.

This would result into a worse performance, specially at python:
https://gitlab.archlinux.org/archlinux/rfcs/-/merge_requests/26

PR #141, opened 2024-03-02
Changing from normal SVT-AV1 to SVT-AV1-PSY

BlueSwordM:
This change serves to improve the visual quality of the video output created by the encoder, as well as improving the its defaults.

I will be the main maintainer for this package as I'm always on the lookout for changes and whatnot.

This will allow a large differentiating factor for CachyOS.

### DIFF
--- a/svt-av1/.SRCINFO
+++ b/svt-av1/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = svt-av1
 	pkgdesc = Scalable Video Technology AV1 encoder and decoder, with added perceptual enhancements for psychovisually optimal AV1 encoding
-	pkgver = 2.0.0.rc2
-	pkgrel = 2
+	pkgver = v2.0.0.rc2.r0.g5cac47c1
+	pkgrel = 1
 	url = https://github.com/gianni-rosato/svt-av1-psy
 	arch = x86_64
 	license = BSD
--- a/svt-av1/.SRCINFO
+++ b/svt-av1/.SRCINFO
@@ -1,8 +1,8 @@
 pkgbase = svt-av1
-	pkgdesc = Scalable Video Technology AV1 encoder and decoder
-	pkgver = 1.8.0
+	pkgdesc = Scalable Video Technology AV1 encoder and decoder, with added perceptual enhancements for psychovisually optimal AV1 encoding
+	pkgver = 2.0.0.rc2
 	pkgrel = 2
-	url = https://gitlab.com/AOMediaCodec/SVT-AV1
+	url = https://github.com/gianni-rosato/svt-av1-psy
 	arch = x86_64
 	license = BSD
 	license = custom: Alliance for Open Media Patent License 1.0
@@ -11,7 +11,7 @@ pkgbase = svt-av1
 	makedepends = nasm
 	makedepends = ninja
 	depends = glibc
-	source = git+https://gitlab.com/AOMediaCodec/SVT-AV1.git#tag=59645eea34e2815b627b8293aa3af254eddd0d69
+	source = git+https://github.com/gianni-rosato/svt-av1-psy.git#tag=5cac47c1eaee9c5229cdf693bf26e90ff6c54b40
 	b2sums = SKIP
 
 pkgname = svt-av1
--- a/svt-av1/PKGBUILD
+++ b/svt-av1/PKGBUILD
@@ -3,7 +3,7 @@
 # Contributor: Thomas Schneider <maxmusterm@gmail.com>
 # Contributor: Zak BlueSwordM <neutronpcxt@gmail.com>
 
-pkgname=svt-av1
+pkgname=svt-av1-psy
 pkgver=v2.0.0.rc2.r0.g5cac47c1
 pkgrel=1
 pkgdesc='Scalable Video Technology AV1 encoder and decoder, with added perceptual enhancements for psychovisually optimal AV1 encoding'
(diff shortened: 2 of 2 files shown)


svt-av1/.SRCINFO
```diff
@@ -11,7 +11,7 @@ pkgbase = svt-av1
 	makedepends = nasm
 	makedepends = ninja
 	depends = glibc
-	source = git+https://gitlab.com/AOMediaCodec/SVT-AV1.git#tag=59645eea34e2815b627b8293aa3af254eddd0d69
+	source = git+https://github.com/gianni-rosato/svt-av1-psy.git#tag=5cac47c1eaee9c5229cdf693bf26e90ff6c54b40
```
lseman, line 14: From checking the repo, I saw that this 5cac47c1eaee9c5229cdf693bf26e90ff6c54b40 is the commit of the RC tag. I don't know if tagging by a commit will work (it may), but it would be better to define as commit, for standard purposes.

svt-av1/PKGBUILD
```diff
@@ -19,22 +20,17 @@ makedepends=(
...
-pkgver() {
-  cd SVT-AV1
-  git describe --tags | sed 's/^v//'
+  sed '/CMAKE_BUILD_TYPE Release/d' -i svt-av1-psy/CMakeLists.txt
 }
 
 build() {
   export LDFLAGS="$LDFLAGS -Wl,-z,noexecstack"
-  cmake -S SVT-AV1 -B build -G Ninja \
+  cmake -S svt-av1-psy -B build -G Ninja \
```
lseman, line 33: Enable AVX512 may be a problem. Either we define two subdirs as in the blas package, or we check for the architecture before building.

lseman (changes requested):
I made some comments, but overall LGTM.

svt-av1/PKGBUILD
vnepogodin: @lseman no, it's not a problem. ENABLE_AVX512 enables building avx512 variants of functions. 
the dispatch takes care which impl to use if cpu supports it

svt-av1/.SRCINFO
BlueSwordM: I will push a fix once I've fixed it.

BlueSwordM: I pushed a fix that tags the proper commit: `svt-av1 v2.0.0.rc2.r0.g5cac47c1-1`

lseman (approved):

gianni-rosato:
Hi, just wanted to give my two cents as the maintainer of SVT-AV1-PSY. First of all, I'm super excited to see this being discussed here!

For those reviewing the PR, some info about [SVT-AV1-PSY](https://github.com/gianni-rosato/svt-av1-psy):
- As you can easily discover, our encoder is focused on being the best AV1 encoder for perceptual fidelity. The reason for this being a fork is that mainline SVT-AV1 has different priorities not necessarily perfectly aligned with maximum visual quality. That being said, we don't intend to compromise on speed or stability to achieve our goals, and so far we can assure we haven't.
- We cannot ensure the same degree of stability as mainline can, but we can get very close. The `master` branch of SVT-AV1-PSY features automated build testing for every commit, and our release cycle is aligned with mainline's (excluding release candidates). We are in touch with a sizeable community of encoders who are constantly using SVT-AV1-PSY and are quick to discover any regressions in the `testing` branch, so things should always be stable upon reaching `master` and afterwards being released.
- SVT-AV1-PSY is a *superset* of SVT-AV1, so any string of command line arguments passed to mainline SvtAv1EncApp should work with our encoder (the binary is of the same name, too). That being said, a command passed to our encoder may not work with mainline due to additional PSY options (e.g `--tune 3`). We won't make any guarantees that bitstreams produced by our encoder and mainline using the same parameters will be identical.

For any others reviewing this PR, I hope this clears things up as how SVT-AV1-PSY would integrate into CachyOS!

ptr1337:
> Hi, just wanted to give my two cents as the maintainer of SVT-AV1-PSY. First of all, I'm super excited to see this being discussed here!
> 
> For those reviewing the PR, some info about [SVT-AV1-PSY](https://github.com/gianni-rosato/svt-av1-psy):
> 
> * As you can easily discover, our encoder is focused on being the best AV1 encoder for perceptual fidelity. The reason for this being a fork is that mainline SVT-AV1 has different priorities not necessarily perfectly aligned with maximum visual quality. That being said, we don't intend to compromise on speed or stability to achieve our goals, and so far we can assure we haven't.
> * We cannot ensure the same degree of stability as mainline can, but we can get very close. The `master` branch of SVT-AV1-PSY features automated build testing for every commit, and our release cycle is aligned with mainline's (excluding release candidates). We are in touch with a sizeable community of encoders who are constantly using SVT-AV1-PSY and are quick to discover any regressions in the `testing` branch, so things should always be stable upon reaching `master` and afterwards being released.
> * SVT-AV1-PSY is a _superset_ of SVT-AV1, so any string of command line arguments passed to mainline SvtAv1EncApp should work with our encoder (the binary is of the same name, too). That being said, a command passed to our encoder may not work with mainline due to additional PSY options (e.g `--tune 3`). We won't make any guarantees that bitstreams produced by our encoder and mainline using the same parameters will be identical.
> 
> For any others reviewing this PR, I hope this clears things up as how SVT-AV1-PSY would integrate into CachyOS!

Thanks gianni!
That did clear things up, but as i have written above, i dont think it is good to replace the default stable version with a -git version.

I really want to avoid using git release in cachyos as default and also replacing offical (maintained) versions of software, with a less tested custom software.
Providing an additonal software, which user can install/replacing them on their own is a better solution.

BlueSwordM:
@vnepogodin I pushed the package name change as requested.

If @ptr1337 accepts, I believe we can push it?
I'm still not sure if it's not better to close this PR and make a new PR with a new PKGBUILD.

That way, the package only gets updated when svt-av1-psy gets updated, rather than with any and every svt-av1 update.

PR #144, opened 2024-03-03
Adding svt-av1-psy as a separate package

BlueSwordM:
Continuation of this PR: https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/141

This time though, it's actually a separate PKGBUILD so that users not only have to manually install it, but it'll get updated separately I believe.

Any improvements welcome.

### DIFF
--- a/svt-av1-psy/.SRCINFO
+++ b/svt-av1-psy/.SRCINFO
@@ -1,4 +1,4 @@
-pkgbase = svt-av1
+pkgbase = svt-av1-psy
 	pkgdesc = Scalable Video Technology AV1 encoder and decoder, with added perceptual enhancements for psychovisually optimal AV1 encoding
 	pkgver = v2.0.0.rc2.r0.g5cac47c1
 	pkgrel = 1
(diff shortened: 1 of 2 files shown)


svt-av1-psy/.SRCINFO
```diff
@@ -0,0 +1,17 @@
+pkgbase = svt-av1
```
ptr1337, line 1: SRCINFO needs to be updated

svt-av1-psy/PKGBUILD
```diff
@@ -0,0 +1,53 @@
...
+  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+build() {
+  export LDFLAGS="$LDFLAGS -Wl,-z,noexecstack"
+  cmake -S svt-av1-psy -B build -G Ninja \
+    -DCMAKE_INSTALL_PREFIX=/usr \
+    -DBUILD_SHARED_LIBS=ON \
+    -DENABLE_AVX512=ON \
+    -DNATIVE=OFF
```
ptr1337, line 44: I think -DNative can be completl removed, since its default off

ptr1337 (changes requested):
LGTM, but check the two comments

svt-av1-psy/PKGBUILD
BlueSwordM: Removed -DNATIVE.

svt-av1-psy/.SRCINFO
BlueSwordM: Updated the SRCINFO.

PR #146, opened 2024-03-05
Systemd boot manager initrd prefix

csutcliff:
Allow the initrd prefix to be set. Default is the current `initramfs`. Useful when using alternatives such as booster. Also check the presence of `mkinitcpio.conf` before greping it as it may not be present if an alternative is being used.


vnepogodin (approved):
LGTM

PR #167, opened 2024-03-25
New package: add mutter-cachyos

ventureoo:
Includes two patches that together and the `CLUTTER_PAINT=disable-dynamic-max-render-time` environment variable should fix the issue of poor external monitor performance on laptops with NVIDIA PRIME. The package is a transparent replacement for mutter and can be installed optionally.


ptr1337:
Could we also include tripple buffering and eventually explicit sync mutter one?

ventureoo:
> Could we also include tripple buffering and eventually explicit sync mutter one?

Triple buffering is already included. With explicit sync I'll take a look.

ventureoo:
> Could we also include tripple buffering and eventually explicit sync mutter one?

Done: https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/167/commits/79b6e492cd3f89847d25029673174b6d6b0044c7. Still need to see how it goes with other patches though.

PR #168, opened 2024-03-25
Allow manual entries for systemd boot to be created without being deleted.

Pogogo007:
Also fixed the remove hook because it was coded to check for a file made by a mhwd-kernel (a manjaro app)


vnepogodin (approved):
LGTM

PR #173, opened 2024-03-31
cachyos-keyring: Provide trusted and revoked

ptr1337:
This adds automatically a trusted and revoked keyring, which we will maybe require in the future, when we go away from a dedicated builder.

This also uses a Makefile for install.
The trusted key is required, since pacman is complaining at the initial import.

I will do some further ISO and calamares testing, before we merge and use this.


sirlucjan (approved):
LGTM

cachyos-keyring/PKGBUILD
```diff
@@ -1,24 +1,26 @@
...
+            'cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e'
+            '4c2677da43b2032453346605cdf869f8c584c6f7d22009cd47f1585fc01c81e91917db237c383dc64ef90baa901a40e887a8a8e693fb97d1851040a16d6b8b62'
+            '8be463fc3c516dcd761ac6d0658605bb688e178339b96d4499d440a9fca8f6b04b2eba6772b8c8b995700d26c734314f12bc386bf824e151e701e583a360c453')
 package() {
-    local destdir=usr/share/pacman/keyrings
-
-    install -d $pkgdir/$destdir
-    install -Dm644 cachyos.gpg  $pkgdir/$destdir/cachyos.gpg
+	cd "${srcdir}"
+	make PREFIX=/usr DESTDIR=${pkgdir} install
```
vnepogodin, line 25: move out of makefile

vnepogodin (changes requested):

cachyos-keyring/PKGBUILD
```diff
@@ -1,24 +1,26 @@
...
 pkgdesc="CachyOS keyring"
 arch=(any)
 url="https://raw.githubusercontent.com/CachyOS/CachyOS-PKGBUILDS/master/$pkgname"
 license=('GPL3')
 groups=(cachyos)
 install=$pkgname.install
-[ -r $install ] || curl -o $install -s --max-time 30 $url/$install
-
 source=(
+  $url/Makefile
```
vnepogodin, line 11: welp why we fetch from net here

vnepogodin (changes requested):

cachyos-keyring/PKGBUILD
ptr1337: I just kept it as it was before, we can remove that

vnepogodin (approved):
LGTM

PR #181, opened 2024-04-09
Added needed files

ripplingsnake:
added proper deckify files and gamescope and 2 dependencies qt5 qt6 tools


vnepogodin (changes requested):
remove built packages

cachyos-deckify/PKGBUILD
```diff
@@ -10,7 +10,7 @@ source=(git+https://github.com/CachyOS/CachyOS-Deckify#commit=e16a15df473da6b74f
 install="$pkgname.install"
 sha256sums=('77e98117921a0c0346c1dd89880132d464448a9825ef69efd915ddfaafc55a82')
 depends=('gamescope' 'mangohud' 'cachyos-gaming-meta' 'gamescope-session-steam-git' 'chimeraos-device-quirks-git'
-         'inputplumber' 'asusctl' 'scx-scheds')
+         'inputplumber' 'asusctl' 'scx-scheds' 'qt5-tools qt6-tools')
```
vnepogodin, line 13: there only `qdbus` is used (which is from qt5-tools). so only qt5-tools needed

tho it would be much better to not pull qt5 without a reason into plasma, so adjust for `qdbus6` in the repo

vnepogodin (changes requested):

ptr1337:
Ive applied with following commit:
https://github.com/CachyOS/CachyOS-PKGBUILDS/commit/c41f94f0e81d0eccb3b701b06693f0fb1276332d

The files wiill be in the other repository adjusted.

PR #188, opened 2024-04-21
add perf to python 3.12

lseman:


ptr1337 (approved):
LGTM, but lets wait to merge this until python 3.12 is in the repo and I would do after that some benchmarks comparing perf vs instrumentation.

Please update checksums in PKGBUILD and .SRCINFO also

PR #226, opened 2024-06-04
nvidia-utils: Enable services for sleep and add config

ptr1337:
Enable these services at the initial installation.
Existing users need to enable the services manually.

Also, set NVReg Path, which fixes graphical corruption on wayland.

### DIFF
--- a/nvidia/nvidia-utils/nvidia-utils.install
+++ b/nvidia/nvidia-utils/nvidia-utils.install
@@ -1,4 +1,3 @@
-SERVICES=("nvidia-suspend" "nvidia-hibernate" "nvidia-resume")
 
 post_upgrade() {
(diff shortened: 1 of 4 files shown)


nvidia/nvidia-utils/nvidia-utils.install
```diff
@@ -6,4 +8,9 @@ post_upgrade() {
 
 post_install() {
   post_upgrade
+
+    echo "Enabling services..."
+    for service in "${SERVICES[@]}"; do
```
ventureoo, line 13: Here we can dispense with loop, just:
```
systemctl enable nvidia-suspend nvidia-hibernate nvidia-resume
```

ventureoo (approved):
LGTM

vnepogodin (approved):

PR #237, opened 2024-06-11
Allow sd-boot manager to source additional config files

Pogogo007:


Pogogo007:
May require more testing

PR #239, opened 2024-06-17
Fix url for wine-cachyos PKGBUILD

h0x32:
Hello, I noticed the URL for wine-cachyos is different than what it actually pulls from, shouldn't it be at least 
https://github.com/CachyOS/wine-cachyos ?


h0x32:
Please check, if this is alright

PR #241, opened 2024-06-18
cachyos-settings: add is-enabled check around uksmd disable

csutcliff:
This prevents an error when it is already disabled or not installed.


ptr1337:
Thanks!

PR #242, opened 2024-06-18
cachyos-settings: add --now to start services immediately

csutcliff:


ptr1337:
I have extra removed this, to start services at runtime to avoid issues. After a cachyos-settings update should be rebooted anyways.

PR #247, opened 2024-06-20
Add trigger rumble support for SDL2

Giovani1906:
Also adds a patch for the Linux API headers in order for SDL to see the trigger rumble parameters.
Needs CachyOS/kernel-patches#64 and WINE/Proton to be compiled against the patched SDL2 in order to work.


Pogogo007:
Out of curiosity which controller supports this? I'm assuming DualSense 5

Giovani1906:
@Pogogo007 It's been tested on a Xbox Series controller.
In theory any controller that has trigger rumble should work. But I don't have a DualSense 5 to test.

1Naim:
I'm going to close this since SDL2 isn't used anymore. Feel free to reopen or open a newer PR if it's still relevant. 

Thanks.

PR #273, opened 2024-07-31
Package opentabletdriver with patch to replace librt-timer with libc-timerfd

1Naim:
Currently in opentabletdriver, a timer expiration is signaled by creating new threads. This whole process is very inefficient and can cause issues with process renicing daemons such as ananicy-cpp. The patch aims to make this more efficient and resolves the conflict with ananicy-cpp caused by the old implementation.

I am also planning to open an issue in [OpenTabletDriver's GitHub](https://github.com/OpenTabletDriver/OpenTabletDriver) to get more attention on this. Meanwhile, I believe we should patch this in as CachyOS OpenTabletDriver users are more susceptible to this bug because ananicy-cpp is enabled by default.

### DIFF
--- a/opentabletdriver/.SRCINFO
+++ b/opentabletdriver/.SRCINFO
@@ -1,0 +1,32 @@
+pkgbase = opentabletdriver
+	pkgdesc = A cross-platform open source tablet driver
+	pkgver = 0.6.4.0
+	pkgrel = 5
+	url = https://opentabletdriver.net
+	install = notes.install
+	arch = x86_64
+	license = LGPL3
+	makedepends = dotnet-sdk>=6.0
+	makedepends = jq
+	makedepends = git
+	depends = dotnet-runtime-6.0
+	depends = gtk3
+	depends = libevdev
+	optdepends = libxrandr: x11 display querying support
+	optdepends = libx11
+	conflicts = digimend-kernel-drivers-dkms-git
+	conflicts = digimend-drivers-git-dkms
+	conflicts = digimend-kernel-drivers-dkms
+	conflicts = digimend-kernel-drivers
+	options = !strip
+	options = !debug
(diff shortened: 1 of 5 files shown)


jamesbt365:
This patch is faulty and causes freezing when using the timer. Please revert as you are now causing problems for OpenTabletDriver devs.

ptr1337:
> This patch is faulty and causes freezing when using the timer. Please revert as you are now causing problems for OpenTabletDriver devs.

Could you provide logs or equal? Currently quite many users are using that, and has been also tested before.
Aditonally this patch has been merged in upstream

jamesbt365:
Report from a user:
image
[L4jIQquWS.json](https://github.com/user-attachments/files/16569651/L4jIQquWS.json)

Interpolator enabled (using native timer).
```json
    {
      "Time": "2024-08-10T08:24:21.9385712-04:00",
      "Group": "Wacom PTK-1240",
      "Message": "Filters: ScuffedInterpolator.ScuffedInterp",
      "StackTrace": null,
      "Level": 1,
      "Notification": false
    },
```

Reproducing it myself:
https://github.com/user-attachments/assets/73d528df-852a-437d-a5bc-5110b54a6977

ptr1337:
Thanks, reverted for now.

jamesbt365:
The original creator of the upstream PR will look at it in a bit but this either needs reverting or patching a second time, which has not been started yet.

1Naim:
I apologize for all the trouble, both to OTD devs and CachyOS devs. I admit that I was a bit too hasty in cherrypicking the patch. On the bright side, a bug was found quicker than it was supposed to. This doesn't change the fact that I should have in fact tested all interpolators before marking this PR mergeable. Very sorry once again.

PR #296, opened 2024-08-19
wine-cachyos: Enable LTO compilation for all native libraries

whrvt:
`lto-fixup.patch` is a rebase of a [patch](https://github.com/openglfreak/wine-tkg-userpatches/blob/df0a89aec388b23a8a82dca6d53e65d25c3a8f86/patches/9999-diffs/lto-fixup.patch) by Torge Matthies, which supersedes [server: Enable link-time optimization](https://github.com/CachyOS/wine-cachyos/commit/cc9ebe323b825db218fb3c257374e81fb5ebc4ed) by allowing all native modules (i.e. modules compiled without mingw) to be compiled with LTO, not just wineserver. 

The relevant compilation flags are added to the PKGBUILD.

### DIFF
--- a/wine-cachyos/.SRCINFO
+++ b/wine-cachyos/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = wine-cachyos
 	pkgdesc = A compatibility layer for running Windows programs, with extra CachyOS flavor
 	pkgver = 9.0.20240831
-	pkgrel = 1
+	pkgrel = 2
 	epoch = 2
 	url = https://github.com/CachyOS/wine-cachyos
 	install = wine.install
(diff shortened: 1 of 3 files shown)


ptr1337:
This is already included.

loathingKernel:
I am positive about using using LTO for everything instead of wineserver only. My issue in the current form is that it modifies the compilation flags in the PKGBUILD. This cannot work for the proton PKGBUILD, it has to be done in Proton's makefile instead. We can keep this open for discussions but we gonna have to apply it independently.

ptr1337:
@whrvt Could you maybe rebase? I would like to test this out on the standalone wine-cachyos, before proceeding with the complete proton.

ptr1337 (approved):
Thanks! I gonna push this today out and we will see, if any bad reports are coming.

ptr1337:
Has been pushed to cachyos-v3, cachyos-v4 and cachyos-znver4

PR #297, opened 2024-08-20
pacutils: Add PrettyProgressBar Config Option

1Naim:
The pacman PKGBUILD we provide has a patch that enables a unicode progress bar when built with the
necessary options. When used with upstream pacutils however, pacconf will print out warnings about
"PrettyProgressBar" config option not existing because of the way it parses configs manually.

For QoL reasons, I have made a patch to add that config option to pacconf directly so it doesn't complain
when used with PrettyProgressBar'ed pacman

### DIFF
--- a/pacman/PKGBUILD
+++ b/pacman/PKGBUILD
@@ -63,6 +63,7 @@ validpgpkeys=(
 
 if $_staging; then
     source+=(78cd453.patch)
+    sha256sums+=('843ad07d4a996ad165f20544c5df5b6deec4496ec30daa38b874bf3fa28afa54')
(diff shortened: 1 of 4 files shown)


ptr1337:
Lets wait for pacman 7 to merge this.

1Naim:
Bump. Pacman 7 has already been merged for a while now. I'm thinking that maybe we should patch in PrettyProgressBar OOTB but leave it disabled as default option. cc @vnepogodin

vnepogodin:
> Bump. Pacman 7 has already been merged for a while now. I'm thinking that maybe we should patch in PrettyProgressBar OOTB but leave it disabled as default option. cc @vnepogodin 

Well we can try that

PR #299, opened 2024-08-22
Add custom gdm package

ventureoo:
This is more needed as a temporary solution until GNOME 47 is released. The package contains two cherry-pick commits to enable Wayland for hybrid graphics with NVIDIA dGPUs, which saves us from having to advise our users to do ``sudo ln -s /dev/null /etc/udev/rules.d/61-gdm.rules``, which is a fairly common issue.


gdm/PKGBUILD
```diff
@@ -0,0 +1,184 @@
...
+END
+
+  install -Dm644 /dev/stdin usr/lib/sysusers.d/gdm.conf <<END
+g gdm 120 -
+u gdm 120 "Gnome Display Manager" /var/lib/gdm
+END
+
+  install -Dm644 /dev/stdin usr/share/glib-2.0/schemas/org.gnome.login-screen.gschema.override <<END
+[org.gnome.login-screen]
+logo='/usr/share/pixmaps/archlinux-logo-text-dark.svg'
```
ptr1337, line 160: Maybe we can override here to CachyOS Logo.

ventureoo: Yeah, sounds good

ventureoo: Hmm, looks like we need add dependency to `cachyos-settings` for this, since `/usr/share/icons/cachyos.svg` is owned by it.

ptr1337: I think we could put then an own package for that?
I think EOS made it like that, so we dont need to maintain all time an own gdm.

ventureoo: > 

Well, we can also put ``/usr/share/glib-2.0/schemas/aa_cachyos.org.gnome.login-screen.gschema.override`` in cachyos-settings

ventureoo: Done with https://github.com/CachyOS/CachyOS-Settings/pull/78

PR #305, opened 2024-08-24
GCC/glibc: Apply ASFLAGS correctly && cherry pick max-page-size from clearlinux

ptr1337:
This adds the LDFLAG -max-page-size=0x1000 (4k) to GCC and glibc.
There was also an issue in applying the ASFLAGS for the x86-64-v4 optimization correctly.

Additionally, we are cherry picking 3 patches from Clearlinux, which are mainly required for the cachyos-v3, cachyos-v4 and cachyos-znver4 repository, due AVX2/512 usage.

In [cachyos-benchmarker](https://github.com/CachyOS/cachyos-benchmarker) there has been an 1,8% improvement found across all benchmarks, using these changes on my 9950X


toolchain/gcc/PKGBUILD
```diff
@@ -121,11 +121,17 @@ build() {
   export CXXLAGS+=" -fno-semantic-interposition"
 
   if [ -n "$_optimize_v3" ]; then
-    export ASFLAGS="-D__AVX__=1 -D__AVX2__=1 -msse2avx -D__FMA__=1 "
-  elif [-n "$_optimize_v4" ]; then
-    export ASFLAGS="-D__AVX__=1 -D__AVX2__=1 -D__AVX512__=1 -msse2avx -D__FMA__=1 "
+    export ASFLAGS+="-D__AVX__=1 -D__AVX2__=1 -msse2avx -D__FMA__=1 "
+    export LDFLAGS+=" -Wl,-z,max-page-size=0x1000"
```
ventureoo, line 125: What is the point of explicitly setting max-page-size? 0x1000 (4kb page) is default value for x86. This would make sense to change if our package had a build condition under ARM arch, where the page size can be changed.

ptr1337: Well, this was cherry picked too from clearlinux. I know, that the page size is default to 4kb.

Maybe glibc/gcc doing some internal things, and make it bigger/smaller here and there? Questionable.
I think having it in wont hurt at all

toolchain/gcc/PKGBUILD
```diff
@@ -121,11 +121,17 @@ build() {
...
+    export CFLAGS+=" -Wl,-z,max-page-size=0x1000"
+    export CXXFLAGS+=" -Wl,-z,max-page-size=0x1000"
```
ventureoo, line 127: Btw, `max-page-size` parameter is only for linker, there is no sense to specify it in CXFLAGS/CXXFLAGS.

toolchain/gcc/PKGBUILD
```diff
@@ -121,11 +121,17 @@ build() {
   export CXXLAGS+=" -fno-semantic-interposition"
 
   if [ -n "$_optimize_v3" ]; then
-    export ASFLAGS="-D__AVX__=1 -D__AVX2__=1 -msse2avx -D__FMA__=1 "
-  elif [-n "$_optimize_v4" ]; then
-    export ASFLAGS="-D__AVX__=1 -D__AVX2__=1 -D__AVX512__=1 -msse2avx -D__FMA__=1 "
+    export ASFLAGS+="-D__AVX__=1 -D__AVX2__=1 -msse2avx -D__FMA__=1 "
```
ventureoo, line 124: Since we are now using `+=`, we need to add a leading space at the beginning so that we don't get into trouble if `ASFLAGS` are already specified somewhere in `makepkg.conf` and don't have a space after them.

ptr1337: right, thanks

toolchain/glibc/PKGBUILD
```diff
@@ -49,17 +51,26 @@ prepare() {
...
 
 }
 
 build() {
 
     if [ -n "$_optimize_v3" ]; then
-      export ASFLAGS="-D__AVX__=1 -D__AVX2__=1 -msse2avx -D__FMA__=1 "
-    elif [-n "$_optimize_v4" ]; then
-      export ASFLAGS="-D__AVX__=1 -D__AVX2__=1 -D__AVX512__=1 -msse2avx -D__FMA__=1 "
+      export ASFLAGS+="-D__AVX__=1 -D__AVX2__=1 -msse2avx -D__FMA__=1 "
```
ventureoo, line 63: Same here

ventureoo:
Otherwise LGTM

toolchain/gcc/PKGBUILD
ptr1337: Yes, I thought about that too - but did not do any tests yet. Im afraid, that this increases generally the memory usage massively.

PR #306, opened 2024-08-24
egl-wayland: 1.1.16

1Naim:


ptr1337:
Mhm, could you push the pkgrel to 1? So that we can remove it, when arch moves it to stable.

ptr1337 (approved):

PR #309, opened 2024-08-26
cachyos-gaming-meta: Remove redundant packages

1Naim:
Removes packages that are either already pulled in by system packages (e.g. pacman, sudo, pipewire, mesa) and packages that are pulled in by dependencies already listed in the meta package (lib32-gtk3)


cachyos-gaming-meta/.SRCINFO
```diff
@@ -1,64 +1,30 @@
...
 	pkgver = 1.0.0
-	pkgrel = 1
+	pkgrel = 2
 	epoch = 1
 	arch = any
-	depends = alsa-lib
 	depends = alsa-plugins
-	depends = dosbox
 	depends = giflib
-	depends = glfw
```
1Naim, line 11: Does anyone know if this is necessary for any games? AFAICT this is just a dependency for `mangohud`

ventureoo: This dependency may be needed for Minecraft if you want to use a system library, such as for native Wayland support on 1.13+.

ventureoo: But I guess it's okay to remove it from depends since mangohud requires it too.

1Naim: > This dependency may be needed for Minecraft if you want to use a system library, such as for native Wayland support on 1.13+.

I'll add it back then, considering `mangohud` is not part of this meta package

ventureoo (approved):
LGTM

PR #317, opened 2024-08-29
systemd-boot-manager: Cleanup series

ventureoo:
Really need testing.


ptr1337:
Tested this so far on my system. No issues found.
Installation testing with ZFS/bcachefs and each also encrypted pending.

ptr1337 (approved):

PR #320, opened 2024-08-29
Add vmware-workstation15 modules with Clang build fixes

ventureoo:
The package probably won't go into the repositories, but it might be worth keeping it for vmware users.


ventureoo:
Updated to the latest vmware-workstation version.

PR #322, opened 2024-08-30
llvm: Build with PGO && ThinLTO

ptr1337:
This enables ThinLTO and PGO for the LLVM packages.

liblllvm.so is a really much used libary, mainly by clang, rust and mesa and improving it with PGO could improve the performance.
Benchmarks and stability testing is pending.


ptr1337:
Using check was successful on my local machine, when building with Clang and ThinLTO.
Arch has a comment to disable LTO, due a failing check, this appears not to happen with the new version or due building with clang.

llvm/PKGBUILD
```diff
@@ -102,25 +102,54 @@ build() {
...
-  cmake .. "${cmake_args[@]}"
+  # Build LLVM with -fprofile generate
+  cmake .. -DCMAKE_BUILD_TYPE=Release -DLLVM_BUILD_INSTRUMENTED=IR -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DLLVM_USE_LINKER=lld -DLLVM_VP_COUNTERS_PER_SITE=6 "${cmake_args[@]}"
+  ninja
+
+  # Run LLVM lit tests to generate profile to be used for PGO
+  ./bin/llvm-lit -v test || true
+
+  # Merge the generated profile
+  llvm-profdata merge -o llvm.profdata "${srcdir}/llvm-18.1.8.src/build/profiles"/*.profraw
```
vnepogodin, line 127: lets use `llvm-$pkgver.src` here instead of hardcoding

llvm/PKGBUILD
```diff
@@ -102,25 +102,54 @@ build() {
...
+
+  # Use Original CFLAGS again and enable ThinLTO + use the profile
+  export PGO_PROFILE="${srcdir}/llvm-18.1.8.src/build/llvm.profdata"
```
vnepogodin, line 130: same

llvm/PKGBUILD
```diff
@@ -102,25 +102,54 @@ build() {
     -DSPHINX_WARNINGS_AS_ERRORS=OFF
   )
 
+  export CC=clang
+  export CXX=clang++
```
vnepogodin, line 106: if we want to use clang as compiler and thin lto, then we would also want to use whole llvm stack

ptr1337:
@1Naim @ventureoo @vnepogodin @sirlucjan 

Testing LLVM package is now available:
```
sudo pacman -U https://archive.cachyos.org/llvm-pgo-thinlto/llvm-18.1.8-6-x86_64_v3.pkg.tar.zst https://archive.cachyos.org/llvm-pgo-thinlto/llvm-libs-18.1.8-6-x86_64_v3.pkg.tar.zst
```

ptr1337:
I have verified the packaging and it looks equal as before. So there is looking all good.
I will plan to do a staged rollout, e.g it will be pushed first to the znver4 repo, together with a mesa rebuild to be freshly linked against.

After some days, and no issues reported we move it further into the other repos.

PR #324, opened 2024-08-30
go: Introduce package for GOAMD=v3/v4

ptr1337:
This adds the go package, for using GOAMD=v3/v4 optimization for the cachyos-v3, cachyos-v4 and cachyos-znver4 repository.


ptr1337:
Compilation successful in chroot. merging

Issue #327, opened 2024-08-31
[RCF] Mesa: Investigate if using Clang with ThinLTO improves the performance


ptr1337:
mesa is currently compiled without LTO due bugs in GCC 14.
We could starting to consider, to enable ThinLTO and a clang built mesa.

Anyone needs to benchmark with AMD or Intel cards, if this improves the performance

ptr1337:
Seems problematic to compile with ThinLTO.
```
ld.lld: error: version script assignment of 'global' to symbol 'driver_descriptor' failed: symbol not defined
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

This can be mitigated with `-Wl,--undefined-version`, also see here:
https://gitlab.freedesktop.org/mesa/mesa/-/issues/8003

ptr1337:
```diff
diff --git a/mesa/mesa/PKGBUILD b/mesa/mesa/PKGBUILD
index b46a1df..39d27c8 100644
--- a/mesa/mesa/PKGBUILD
+++ b/mesa/mesa/PKGBUILD
@@ -214,10 +214,23 @@ build() {
     -D vulkan-layers=device-select,intel-nullhw,overlay
   )
 
+  # Set Clang as compiler
+  export AR=llvm-ar
+  export CC=clang
+  export CXX=clang++
+  export NM=llvm-nm
+  export RANLIB=llvm-ranlib
+
+  export CFLAGS+=" -flto=thin"
+  export CXXFLAGS+=" -flto=thin"
+  export LDFLAGS+=" -Wl,--undefined-version -fuse-ld=lld"
   # Build only minimal debug info to reduce size
   CFLAGS+=" -g1"
   CXXFLAGS+=" -g1"
 
+  # LTO needs more open files
+  ulimit -n 4096
+
   # Inject subproject packages
   export MESON_PACKAGE_CACHE_DIR="$srcdir"
 
```

Diff to get ThinLTO correclty working. WIll provide some test packages

ptr1337:
@1Naim @SoulHarsh007 
Testing Packages can be found here:
https://archive.cachyos.org/mesa-thinlto/

1Naim:
With a Radeon 660M (iGPU), I did not notice any performance gains from ThinLTO. However, there is a sizeable reduction in package size (notably in lib32-mesa, but approx. 28MB total). 

[...]

PR #328, opened 2024-08-31
mesa/lib32-mesa: Use Clang and ThinLTO

ptr1337:
Performance and Stability testing pending

More info:
https://github.com/CachyOS/CachyOS-PKGBUILDS/issues/327


1Naim:
I've been using this for 3 days now, and while I can say I haven't faced any issues with this package, there are any performance improvements by using ThinLTO from the benchmarks (vkmark and RE6 Bench) and usage that I've done. 

IMO if we want to improve performance in the graphics stack, we should instead make changes to the codebase, e.g. backporting some open MRs that have tangible performance benefits, e.g. [mesa/!25576](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25576)

ptr1337:
Closing, since it does not provide a real benefit currently.

PR #331, opened 2024-09-01
cachyos-rate-mirrors: Use the CachyOS CDN as first mirror after ranking

ptr1337:
The CDN provides a updated archlinux mirror, which gets synced pretty frequently.
Also, most users have quite good download speed on these.

There are not too many packages pulled from this repo, therefore lets put it on the top as default after ranking.

We have been using this mirror as default for the installation and there were never real complains or issues with it at all too.


SoulHarsh007 (approved):
LGTM!

vnepogodin (approved):
legit

PR #332, opened 2024-09-02
feat: Add zlib-ng

### LINKED ISSUE #330: [RFC] Consider to replace zlib with zlib-ng
Copy from Fedora RFC:

Zlib has long served as a reliable compression library, but its performance on modern CPU architectures does not fully utilize their capabilities. The transition to Zlib-ng aims to enhance the compression efficiency and performance for Fedora while ensuring compatibility with existing packages and libraries.

## Benefits for CachyOS

- Zlib-ng supports hardware acceleration when available
- Adler32 implementation using SSSE3, AVX2, AVX512, AVX512-VNNI, Neon, VMX & VSX
- CRC32-B implementation using PCLMULQDQ, VPCLMULQDQ, ACLE, & IBM Z
- Hash table implementation using CRC32-C intrinsics on x86 and ARM
- Slide hash implementations using SSE2, AVX2, ARMv6, Neon, VMX & VSX
- Compare256 implementations using SSE2, AVX2, Neon, POWER9 & RVV
- Inflate chunk copying using SSE2, SSSE3, AVX, Neon & VSX
- Support for hardware-accelerated deflate using IBM Z DFLTCC
- Includes Intel and Cloudflare optimizations
- Uses code sanitizers, fuzzing, and code coverage

## Integration

1. Install zlib-ng to users system
2. Provide and install zlib-ng-compat from the AUR into the cachyos-repository
3. Likely, create a lib32-zlib-ng and lib32-zlib-ng compat package, which replaces zlib-ng

## Question
We need to look into, if there are any recompilation required. Normally, this **should** work as drop in replacement but should be looked deeper into

1Naim:
We aim to replace zlib with zlib-ng to make use of the optimizations from the next generation implementation

Some benchmarks to highlight the benefits of this replacement:
- [zlib-ng - 2.0.0 Benchmark Discussions](https://github.com/zlib-ng/zlib-ng/discussions/871)
- [AWS - Comparisons between zlib-cloudflare, zlib-ng and other zlib forks](https://aws.amazon.com/blogs/opensource/improving-zlib-cloudflare-and-comparing-performance-with-other-zlib-forks/)
- [Megastorm's Systems zlib-ng vs zlib](https://www.megastormsystems.com/news/benchmarks-results-of-zlib-and-zlib-ng-running-on-some-amd-and-intel-cpus)

Closes #330

### DIFF
--- a/zlib-ng/zlib-ng/PKGBUILD
+++ b/zlib-ng/zlib-ng/PKGBUILD
@@ -1,6 +1,12 @@
+### Arch Linux PKGBUILD Maintainers:
 # Maintainer: Levente Polyak <anthraxx[at]archlinux[dot]org>
 # Contributor: Chocobo1 <chocobo1 AT archlinux DOT net>
 # Contributor: Jacek Szafarkiewicz <szafar at linux dot pl>
+### CachyOS PKGBUILD Maintainers:
+# Maintainer: Eric Naim <dnaim@cachyos.org>
+# Maintainer: Peter Jung <admin@ptr1337.dev>
+
+
 pkgbase=zlib-ng
 pkgname=(
   $pkgbase
--- a/zlib-ng/zlib-ng/PKGBUILD
+++ b/zlib-ng/zlib-ng/PKGBUILD
@@ -1,0 +1,82 @@
+# Maintainer: Levente Polyak <anthraxx[at]archlinux[dot]org>
+# Contributor: Chocobo1 <chocobo1 AT archlinux DOT net>
+# Contributor: Jacek Szafarkiewicz <szafar at linux dot pl>
+pkgbase=zlib-ng
+pkgname=(
+  $pkgbase
+  $pkgbase-compat
+)
+pkgver=2.2.1
+pkgrel=2
+pkgdesc='zlib replacement with optimizations for next generation systems'
+url='https://github.com/zlib-ng/zlib-ng'
+arch=('x86_64')
+license=('custom:zlib')
+depends=(
+  glibc
+)
+makedepends=(
+  cmake
+  ninja
+)
+source=("${url}/archive/refs/tags/$pkgver/zlib-ng-${pkgver}.tar.gz")
+sha256sums=('ec6a76169d4214e2e8b737e0850ba4acb806c69eeace6240ed4481b9f5c57cdf')
+b2sums=('eacd6e01b8792d1646054b32274c486021fa3c4341a5ce4e19d7afcccf0fc8645a20e78aedad2b353d662b2b82641780c55ff8e4e255c3fe9ffb67fce2001552')
+
+
+build() {
+  cd "zlib-ng-${pkgver}"
+
+	# WITH_UNALIGNED - unaligned access invokes undefined behaviour,
+	#   see https://github.com/gentoo/gentoo/pull/17167 for more info.
+  local _options=(
+    -G Ninja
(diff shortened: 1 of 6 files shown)


ptr1337:
We also need to think about, how we do the integration.

e.g we would need to provide a replace=(zlib) for example, so that user will get when updating automatically the update. But for the testing right now that fits

1Naim:
Currently `zlib-ng-compat` conflicts with `zlib` so for any users that currently build this manually they will get prompted.

It is true however that when we actually get down to merging this, we need to add that as an option.

ptr1337:
We maybe want to add before the build part:
```
export CFLAGS+=" -fno-semantic-interposition"
export CXXLAGS+=" -fno-semantic-interposition"
```

ptr1337:
Clearlinux uses following to PGO the package:
```
cat *.c | ./minigzip -6 | ./minigzip -d > /dev/null
cat *.c | ./minigzip -4 | ./minigzip -d > /dev/null
cat *.c | ./minigzip -9 | ./minigzip -d > /dev/null
cat minigzip | ./minigzip -6 | ./minigzip -d > /dev/null
cat minigzip | ./minigzip -4 | ./minigzip -d > /dev/null
cat minigzip | ./minigzip -9 | ./minigzip -d > /dev/null
```

Since mini(g)zip is provided by zlib. We could still run this as PGO profiling, due the compact package (likely). Needs to be tested.

Following C/CXX Flags should be used for profiling:
`-fprofile-generate -fprofile-update=atomic`
`-fprofile-use -fprofile-dir=$PGODIR -fprofile-correction`

sirlucjan:
Ive compiled and installed it - without any problem.

zlib-ng/zlib-ng/PKGBUILD
```diff
@@ -0,0 +1,88 @@
...
+  cd "zlib-ng-${pkgver}"
+  DESTDIR="${pkgdir}" cmake --install build
+  install -Dm 644 LICENSE.md -t "${pkgdir}/usr/share/licenses/${pkgname}"
+  install -Dm 644 README.md -t "${pkgdir}/usr/share/doc/${pkgname}"
+}
+
+package_zlib-ng-compat() {
+  pkgdesc+=" (zlib compat)"
+  provides=('zlib' 'libz.so')
+  replaces=('zlib')
```
1Naim, line 80: This need a bit more evaluation if we want to actually ship it to all users using our repo. 

There aren't any issues on my side so I'm +1 mostly.

sirlucjan:
I'm testing, using and it works.

zlib-ng/zlib-ng/PKGBUILD
```diff
@@ -0,0 +1,91 @@
...
+
+
+build() {
+  cd "zlib-ng-${pkgver}"
+
+  export CFLAGS+=" -fno-semantic-interposition"
+  export CXXLAGS+=" -fno-semantic-interposition"
+
+	# WITH_UNALIGNED - unaligned access invokes undefined behaviour,
+	#   see https://github.com/gentoo/gentoo/pull/17167 for more info.
```
ptr1337, line 40: I think the formatting here is a bit failed

ptr1337 (changes requested):

ptr1337 (approved):
So, lets merge this for now and put it into the cachyos-testing-v3 repository, and let some people test.

After that, we can try to push it to stable.

PR #333, opened 2024-09-02
feat: Use proper PKGBUILD for mpv-git

1Naim:
Currently the mpv-git provided in our repositories is using the PKGBUILD directly from AUR. The PKGBUILD from AUR is not meant to be distributable, since the PKGBUILD enables features and finds dependencies based on the build system. Due to this, the AUR PKGBUILD is more suitable for building manually for each user.

As a solution, I have adopted to use the PKGBUILD by Arch Linux and adjusted it to pull the source from git instead of a tagged release.

Using a custom PKGBUILD can open new doors to maybe some optimizations that can be possible, e.g. PGO or clang-LTO


ptr1337:
Mhm. Im not sure about, since we would need to manually update it then, but Ill take a watch that we integrate into the weekly -git workflow.

ptr1337 (approved):

PR #355, opened 2024-09-25
Drop winesync from proton-cachyos and from repos

1Naim:
We do not ship winesync inside proton-cachyos anymore. As it is also essentially abandonware and can also cause issues, drop it from the repo too.


loathingKernel:
LGTM

loathingKernel (approved):

PR #363, opened 2024-09-29
Update zsh

Dekomoro:
I updated zsh, dropped the asc signature from the pkgbuild so you don't need to apply --skippgpcheck when building and synced the patches with arch linux(including renaming the autocompletion fix).
I adjusted the maintainer status since it appears ptr wanted me to maintain.

Let me know if there are any issues.


zsh/PKGBUILD
```diff
@@ -1,39 +1,56 @@
...
 pkgbase=zsh
 pkgname=('zsh')
 pkgver=5.9
-pkgrel=4
+pkgrel=6
 arch=('x86_64')
 url='https://www.zsh.org/'
 license=('custom')
 makedepends=('pcre' 'libcap' 'gdbm' 'yodl' 'gcc')
-source=("https://www.zsh.org/pub/zsh-${pkgver}"{,-doc}".tar.xz"{,.asc}
```
1Naim, line 12: Please do not remove the .asc file. See [PKGBUILD/Integrity](https://wiki.archlinux.org/title/PKGBUILD#Integrity) for details but the takeaway is that it is good practice to provide the digital signatures that upstream provides when distributing PKGBUILDs

zsh/PKGBUILD
```diff
@@ -1,39 +1,56 @@
...
 
-    # 50629: do not use egrep in tests
-    patch -Np1 <../0001-50629-do-not-use-egrep-in-tests.patch
-
-    # https://github.com/zsh-users/zsh/commit/3e3cfabcc74dc79d4d8717c4e5859d8d01be6c54
-    # https://www.zsh.org/mla/workers/2022/msg00652.html
-    patch -Np1 <../fix-autocompletion.patch
+	# 50629: do not use egrep in tests
+	# 50641: use 'int main()' in test C-codes in configure
+	patch -Np1 < ../0001-50629-do-not-use-egrep-in-tests.patch
```
1Naim, line 36: Please fix the identation here.

1Naim (changes requested):

zsh/PKGBUILD
```diff
@@ -1,39 +1,56 @@
...
 url='https://www.zsh.org/'
 license=('custom')
 makedepends=('pcre' 'libcap' 'gdbm' 'yodl' 'gcc')
-source=("https://www.zsh.org/pub/zsh-${pkgver}"{,-doc}".tar.xz"{,.asc}
-    '0001-50629-do-not-use-egrep-in-tests.patch'
-    'fix-autocompletion.patch'
-    'zprofile')
+source=("https://www.zsh.org/pub/zsh-${pkgver}"{,-doc}".tar.xz"
+        '0001-50629-do-not-use-egrep-in-tests.patch'
+        '0001-50641-use-int-main.patch'
```
vnepogodin, line 15: Can you provide remote patches instead of local ones? Would be more preferable if possible

1Naim: Local patches are also used from Arch

vnepogodin: we also local patches, but remote ones still more preferable)

1Naim: I don't know if that's possible considering zsh's development is through mailing list.

vnepogodin: Yeah we can add link to the commit from their github mirror or so

Dekomoro: It appears the patches to not cleanly apply if I grab the mirrored versions, is it possible arch is using rebased patches?

zsh/PKGBUILD
Dekomoro: it is now not removed, would you prefer that I restore the validpgpkeys value as well?

zsh/PKGBUILD
```diff
@@ -1,39 +1,60 @@
...
-    patch -Np1 <../0001-50629-do-not-use-egrep-in-tests.patch
-
-    # https://github.com/zsh-users/zsh/commit/3e3cfabcc74dc79d4d8717c4e5859d8d01be6c54
-    # https://www.zsh.org/mla/workers/2022/msg00652.html
-    patch -Np1 <../fix-autocompletion.patch
+    # 50641: use 'int main()' in test C-codes in configure
+    patch -Np1 < ../0001-50629-do-not-use-egrep-in-tests.patch
+    patch -Np1 < ../0001-50641-use-int-main.patch
+    
+   	# 50325: revert 38150 and fix in calling function cfp_matcher_range() instead
```
1Naim, line 43: Here too

zsh/PKGBUILD
1Naim: Yes.

zsh/PKGBUILD
1Naim: I'll look into this in a bit.

zsh/PKGBUILD
Dekomoro: weird, that wasn't appearing locally, should be fixed now though

zsh/PKGBUILD
Dekomoro: it appears by looking at the diff of the arch patches and the files are slightly different, is it possible that for example that the link here is the wrong way to get the patch? https://github.com/zsh-users/zsh/commit/4fc5dc0292acd77f17281f451774ba2ca4203026.patch

1Naim: It's not wrong. If it's slightly different then Arch must have rebased the patches so it applies on top of 5.9 instead of master which is what the patch is based on.

zsh/PKGBUILD
Dekomoro: I restored this

zsh/PKGBUILD
1Naim: I see. Patches are failing to apply in the changelog. The actual code itself patches cleanly. We can leave it as local patches thank you.

1Naim (approved):
LGTM

vnepogodin:
missed to update .SRCINFO after modification to PKBUILD

Dekomoro:
> missed to update .SRCINFO after modification to PKBUILD

done

vnepogodin (approved):
LGTM

PR #374, opened 2024-10-08
kwin: 6.2.0 + mouse acceleration fix

1Naim:
kwin 6.2.0 removes the need for the tearing and vrr patch we cherrypicked. Copy-paste fix is also already merged in 6.2.0 so its removed. I've cherrypicked a commit thats queued for 6.2.1 that fixes mouse acceleration when the cursor moves too fast.


1Naim:
Built successfully.

PR #377, opened 2024-10-09
pacman Build with PrettyProgressBar

1Naim:
Builds pacman with PrettyProgressBar functionality. This option, when enabled, will render the progress bar using
specifing unicode characters so it looks "pretty". It is disabled by default for compatibility reasons (some terminals
cannot render unicode characters nicely, and not all fonts have those specific unicode characters).

Currently, I have just flipped ``$_staging` to true for easy disabling of the patch, but this may change later.

To complement `PrettyProgressBar` enablement in pacman. I have also added `pacutils` package that adds `PrettyProgressBar` as a valid config option so it doesn't spam warnings.


vnepogodin:
we use staging flag for building against staging branch. Currently its the same and doesnt build from staging.

Regarding patch the "disabling" of it can be done via commenting it out, because its automatically patched if present in source array

1Naim:
> Regarding patch the "disabling" of it can be done via commenting it out, because its automatically patched if present in source array

That's a good point. I also wasn't aware about building from staging part as that wasn't apparent with the current state of the PKGBUILD. I've updated the PR accordingly tho, thanks for the pointers ^^

vnepogodin (approved):

PR #380, opened 2024-10-11
Drop unmaintained packages

1Naim:
A bit of cleaning in the repos, I've dropped most packages that haven't seen any updates in years or are unmaintained. 

I've left most of the WM/DE settings in since we provide those environments in install but please comment if it's unmaintained so I can remove them in this PR too.

A few other things I've left untouched, let me know if these are still maintained or not so I can respond accordingly:
- https://github.com/CachyOS/char-white.git
- https://gitlab.com/nycko123/pak

Let me know if there are packages that I've missed or removed a still-maintained package so I can add or remove them


vnepogodin (approved):
legit

PR #381, opened 2024-10-12
cachyos-ananicy-rules: Use HEAD instead of specific commits

1Naim:
This avoids the need of bumping the commit everytime there's an update and updating the PKGBUILD for it. We can also integrate this to https://github.com/CachyOS/PKGBUILDS-git/ for less maintenance burden.

Merge with https://github.com/CachyOS/PKGBUILDS-git/pull/1


ptr1337:
The bumping is intended to have verified commit hashes

PR #385, opened 2024-10-16
[pacman] Add commented ILoveCandy

TheWyn:
PrettyProgressBar patch includes an option for ILoveCandy and isn't included in the config.  All the other options are already added here.


1Naim:
ILoveCandy is part of core pacman but yeah we can add this to the makepkg.conf

1Naim (approved):

ptr1337:
Why do we want this? This is actually already set in the install pacman config:
https://github.com/CachyOS/CachyOS-Live-ISO/blob/master/archiso/airootfs/etc/pacman-more.conf#L35

1Naim:
My thinking was mainly for posterity in the pacman.conf provided by the package itself, considering all the other options have the comments too.

TheWyn:
It is not included in the .pacnew whenever pacman has an update

ptr1337:
Thats an argument

ptr1337 (approved):

1Naim (changes requested):
This flew under my radar, but please update the checksums too. You can do this with `updpkgsums` and `makepkg --printsrcinfo > .SRCINFO` to update .SRCINFO

vnepogodin (approved):

1Naim (approved):
Thanks!

PR #412, opened 2024-11-10
egl-wayland: Remove local ICD json

1Naim:
Since https://github.com/NVIDIA/egl-wayland/commit/954e094faacdacad6d82905d2f0a674c3e37fbec, there is no need now to use a locally provided json file since it is now provided from upstream. Since these are identical there is also no need to bump pkgrel and do a rebuild


ptr1337:
Can you contribute that to arch?

PR #419, opened 2024-11-14
chwd: Cherrypick amd profile update due recent mesa changes

1Naim:
This should also fix chwd failing in ISO installs since it won't fetch the obsolete packages anymore.


vnepogodin (approved):
LGTM

PR #420, opened 2024-11-14
Disable RTLD_DEEPBIND on all PHP SAPIs except for apache

danog:
This pull request changes the default behavior of dlopen to use RTLD_DEEPBIND only when compiling the Apache SAPI.

This fixes compatibility with custom allocators like jemalloc on newer glibc versions, while still retaining the old DEEPBIND behavior for apache, which seems to have been the [reason why it was added](https://github.com/php/php-src/issues/10670#issuecomment-1503458109).

Also see https://github.com/php/php-src/issues/10670, https://github.com/php/php-src/pull/11094, https://github.com/php/php-src/pull/16779


ptr1337 (approved):
Thank you! Looks good to me.

Will push it to the repository.

danog:
@ptr1337 Thanks, mind rebuilding the package?

Issue #337, opened 2024-09-09
[cachy-browser] Allow replacing of jemalloc with hardened_malloc


KurtFuchs:
As described in [RedHat Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2260766), it is possible to allow Firefox to use the hardened_malloc implementation (needs to be installed on the system) when using  ```ac_add_options --enable-replace-malloc``` while still being able to Fallback to jemalloc. Fedora already uses the flag with the standard package. This would help improve the security of Cachy Browser. [For information on hardened_malloc ](https://github.com/GrapheneOS/hardened_malloc).

PR #423, opened 2024-11-16
cachy-browser: Rework PKGBUILD and drop librewolf stuff

ptr1337:
This reworks the cachy-browser PKGBUILD and patches.

We are dropping the librewolf patches and go for a "tuned" firefox version, which provides a more secure and faster config.
The gentoo patches will be still applied.

Generally, this is done since librewolf does update their patches in a very slow cycle and specially if CVE's or equal are present, we can not push when we want.

Drawsbacks:
- Loosing the "Cachy" Page for various setting
- Some less security directly in the patches

Im  also converting the PKGBUILD to follow the upstream archlinux PKGBUILD, so that this keeps same as possible.
Currently we can not compile cachy-browser for higher architectures, because LLVM 18 has a bug.

This will be fixed as soon archlinux does push LLVM19 into the repository.
Outisde of that, we are using only ThinLTO instead of "cross,full", which speeds up the compilation, while having barely performance looses.
Fixes: https://github.com/CachyOS/CachyOS-PKGBUILDS/issues/337
Signed-off-by: Peter Jung <admin@ptr1337.dev>


sirlucjan (approved):
Tested-by: Piotr Gorski [piotrgorski@cachyos.org](mailto:piotrgorski@cachyos.org)

ventureoo:
Can we accomplish https://github.com/CachyOS/CachyOS-PKGBUILDS/issues/337 in this case?

ptr1337:
> Can we accomplish #337 in this case?

Done

ventureoo:
> > Can we accomplish #337 in this case?
> 
> Done

Thanks, can also be labeled as `Fixes: https://github.com/CachyOS/CachyOS-PKGBUILDS/issues/337`.

PR #424, opened 2024-11-17
Add qtcurve

TheWyn:
Qt6 fork of qtcurve to use in Application Style in Kde Plasma

image


ptr1337:
Is this PKGBUILD different from the AUR one?
If not, then I would just put it via the AUR pkg

TheWyn:
> Is this PKGBUILD different from the AUR one? If not, then I would just put it via the AUR pkg

Yes, the aur package doesn't compile.

PR #426, opened 2024-11-17
Drop Legacy NVIDIA Drivers

1Naim:
470 drivers has gone EOL since end of September. Considering that these haven't been updated in over 5 months, we should just
drop it. It probably won't work with newer kernels anyway (case in point: latest drivers don't work on 6.12 without patching). We don't need to talk about 390xx drivers.

We should also drop support in chwd eventually.


vnepogodin:
Well that would be great from maintainer perspective, but not from user. Although we can do nothing until it requires little changes

ventureoo:
I had no intention to drop 470.xx until NVK provided the promised Kepler support.

ventureoo:
Honestly it's not a big deal for me to maintain these packages for a while longer.

1Naim:
> Honestly it's not a big deal for me to maintain these packages for a while longer.

Alright then. My main concern was noone maintaining these. The other was checking if these even still work with the latest kernels. I'll close this PR, thank you.

ptr1337:
> > Honestly it's not a big deal for me to maintain these packages for a while longer.
> 
> Alright then. My main concern was noone maintaining these. The other was checking if these even still work with the latest kernels. I'll close this PR, thank you.

The problem we have is maintaining upstream kernel support (e.g latest).

We should maybe install maybe the lts kernel with as default.

PR #434, opened 2024-11-21
bpftune-git: Rework PKGBUILD

1Naim:
- Uses libdir to install to /lib/ instead of using sed
- Rework fix-makefile.patch to instead just remove march (should respect makepkg.conf's) and use O3 flag

Oracle states that there will be a proper tagged release soon.


ptr1337 (approved):
LGTM, thanks

PR #435, opened 2024-11-21
mesa: 24.3.0-1

ptr1337:
Updated for mesa 24.3.0 and needs some testing.
We maybe want to also wait for archlinux to update their PKGBUILD and then sync it. Compilation was successful and the size got around 10mb lowers each - maybe because the gallium-omx driver was removed.


1Naim (approved):
Tested-by

PR #445, opened 2024-12-04
opentabletdriver: Sync with AUR

1Naim:


opentabletdriver/PKGBUILD
```diff
@@ -6,32 +6,38 @@
...
-pkgrel=7
+pkgrel=8
 pkgdesc="A cross-platform open source tablet driver"
 arch=('x86_64')
 url="https://opentabletdriver.net"
-license=('LGPL3')
+
+# license FIXME tracked upstream:
+# https://github.com/OpenTabletDriver/OpenTabletDriver/issues/3571
+license=('LGPL3') # FIXME: use SPDX format
```
ptr1337, line 17: Please put your SPDX Format

ptr1337 (changes requested):

opentabletdriver/PKGBUILD
1Naim: The proper identifier hasn't been decided yet by the OTD team. I believe that the next point release will be released some time next week, so we can drop this by then. There doesn't seem to be any updates other than some cleanups in this sync so I'll just close this.

PR #448, opened 2024-12-05
mesa: 24.3.1

1Naim:
Removed the b2sums array. Refer to commit log for rationale. It's controversial so let me know if you don't want this.

### DIFF
--- a/mesa/lib32-mesa/.SRCINFO
+++ b/mesa/lib32-mesa/.SRCINFO
@@ -1,6 +1,6 @@
 pkgbase = lib32-mesa
 	pkgdesc = Open-source OpenGL drivers - 32-bit
-	pkgver = 24.3.0
+	pkgver = 24.3.1
 	pkgrel = 2
 	epoch = 1
 	url = https://www.mesa3d.org/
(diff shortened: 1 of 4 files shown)


ptr1337:
Removing sha256 would be better and keeping b2 :)

1Naim:
> Removing sha256 would be better and keeping b2 :)

Any reasons for this? https://wiki.archlinux.org/title/PKGBUILD#Integrity:

> The checksum type and values should always be those provided by upstream, such as in release announcements.

As far as I can see in https://docs.mesa3d.org/relnotes/24.3.1.html, mesa doesn't provide b2 checksums. Although now that I see it again, we should probably use sha512 instead since that's preferred over sha256.

vnepogodin:
> > Removing sha256 would be better and keeping b2 :)
> 
> Any reasons for this? https://wiki.archlinux.org/title/PKGBUILD#Integrity:
> 
> > The checksum type and values should always be those provided by upstream, such as in release announcements.
> 
> As far as I can see in https://docs.mesa3d.org/relnotes/24.3.1.html, mesa doesn't provide b2 checksums. Although now that I see it again, we should probably use sha512 instead since that's preferred over sha256.

Weää I think we should keep as arch there

1Naim:
> > > Removing sha256 would be better and keeping b2 :)
> > 
> > 
> > Any reasons for this? https://wiki.archlinux.org/title/PKGBUILD#Integrity:
> > > The checksum type and values should always be those provided by upstream, such as in release announcements.
> > 
> > 
> > As far as I can see in https://docs.mesa3d.org/relnotes/24.3.1.html, mesa doesn't provide b2 checksums. Although now that I see it again, we should probably use sha512 instead since that's preferred over sha256.
> 
> Well I think we should keep as arch there

Sure, I'll just drop the last commit then so it's not blocking the version update.

ptr1337 (approved):

PR #457, opened 2024-12-13
mesa: Fix various issues with Indiana Jones

1Naim:


mesa/mesa/.SRCINFO
```diff
@@ -1,7 +1,7 @@
 pkgbase = mesa
 	pkgdesc = Open-source OpenGL drivers
 	pkgver = 24.3.1
-	pkgrel = 4
+	pkgrel = 5
```
ptr1337, line 4: Could you bump to 6? Ive failed once in znver4 repo, and then we have it equal again.

Outside of that fine

ptr1337 (changes requested):

ptr1337 (approved):

PR #461, opened 2024-12-21
kwin: Implement fifo-v1 wayland protocol

1Naim:
Allows the client to get proper FIFO behaviour, by waiting for vblank instead of frame callbacks.
This is the last remaining bit needed to get fifo-v1 in the desktop, as both mesa stable and main contains the necessary
changes to use fifo-v1, with or without commit-timing-v1.


ptr1337:
This does depend on an extra mesa patch to work independent from commit thing protocol?

1Naim:
> This does depend on an extra mesa patch to work independent from commit thing protocol?

That has already been merged as of 24.3.2, so no extra patches are needed.

ptr1337 (approved):

PR #463, opened 2024-12-28
pango: Build with PGO for 1.10x speedup

joebonrichie:
Test File
---------
https://raw.githubusercontent.com/joebonrichie/benchmarks/refs/heads/master/pango-layout-profile.c

Benchmark
---------
hyperfine "./pango-layout-profile"
Before: 12.823 s | Now: 11.569 s

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,89 @@
+# Maintainer: Jan Alexander Steffens (heftig) <heftig@archlinux.org>
+# Contributor: Jan de Groot <jgc@archlinux.org>
+# Contributor: Joey Riches (joebonrichie) <josephriches@gmail.com>
+
+pkgbase=pango
+pkgname=(
+  pango
+  pango-docs
+)
+pkgver=1.54.0
+pkgrel=2
+epoch=1
+pkgdesc="A library for layout and rendering of text"
+url="https://www.pango.org/"
+arch=(x86_64)
+license=(LGPL-2.1-or-later)
+depends=(
+  cairo
+  fontconfig
+  freetype2
+  fribidi
+  glib2
+  glibc
+  harfbuzz
+  libsysprof-capture
+  libthai
+  libx11
+  libxft
+  libxrender
+)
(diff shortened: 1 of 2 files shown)


ptr1337:
Hey!

Thanks for the PR. I will do some testing and then will push it to the repository. PKGBUILD is looking good.



> I can't actually tell whether cachy uses -O3 by default for it's own packages it provides itself over arch repos due to O3 being present in makepkg.conf. However, this in case, it causes in a performance regression so O2 is explicitly appended to build flags.
> 
> Package was benchmarked against cachyos-extra-v3/pango.

Yes, we are using currently globally -O3 - nice catch that it has a regression.

PKGBUILD
```diff
@@ -0,0 +1,88 @@
+# Maintainer: Jan Alexander Steffens (heftig) <heftig@archlinux.org>
+# Contributor: Jan de Groot <jgc@archlinux.org>
+
+pkgbase=pango
+pkgname=(
+  pango
+  pango-docs
+)
+pkgver=1.54.0
+pkgrel=1
```
1Naim, line 10: Please bump this to 2, so it correctly replaces the existing package in `extra` and `cachyos-extra-*`

PKGBUILD
```diff
@@ -0,0 +1,88 @@
...
+    -D documentation=true
+    -D sysprof=enabled
+  )
+
+  export PROFILE_DATA="${srcdir}/PGO"
+  # O3 causes a performance regression in tested benchmarks, ensure we use O2
+  export CFLAGS="${CFLAGS} -fprofile-generate -fprofile-update=atomic -fprofile-dir=$PROFILE_DATA -O2"
+
+  arch-meson pango build "${meson_options[@]}"
+  meson compile -C build -v
```
1Naim, line 56: Is there any reason you're using `-v` flag here? I'm not necessarily against, I'm just wondering why.

ptr1337 (approved):

PR #468, opened 2025-01-02
[kde-settings] Move depend package to optional

TheWyn:
Konsole is installed by default and Alacritty itself is optional in the installer.  No need to store config files for a program that likely isn't installed.  Make Alacritty depend the config not the DE config.

### DIFF
--- a/cachyos-kde-settings/PKGBUILD
+++ b/cachyos-kde-settings/PKGBUILD
@@ -15,7 +15,9 @@ depends=('ttf-fantasque-nerd'
     'noto-fonts'
     'ttf-fira-sans'
     'capitaine-cursors'
-    'cachyos-alacritty-config'
+)
+optdepends=(
+  'cachyos-alacritty-config'
 )
 install=$pkgname.install
 provides=('cachyos-desktop-settings')


vnepogodin:
I don't see much sense into moving that package into optionals, would make more sense to just remove it from depends at all (as it will never be installed).

"settings"(which provide files in /etc/skel) packages have the only purpose, is to install configs during install.

About Konsole and Alacritty: alacritty was "default" terminal emulator on cachyos(if you can say like that), and as kde config is basically "vanilla" config Konsole is used.

vnepogodin:
Alacritty settings package was added into the depends to provide drop-in replacement for config which was already in the kde settings package.

TheWyn:
It'd be better to add Alacritty to the repo and add the config package as a depend.  Current method adds upkeep keeping it in every DE config and regardless if Alacritty is installed or not.

>Regarding Konsole and Alacritty: Alacritty was the default terminal emulator on CachyOS (if you can call it that), and since the KDE configuration is basically "vanilla", Konsole is used.

Wasn't that also due to multiple DE options for the live iso download?

vnepogodin:
> It'd be better to add Alacritty to the repo and add the config package as a depend. Current method adds upkeep keeping it in every DE config and regardless if Alacritty is installed or not.
> 

Well it's not better, it adds mess into the repository(and "Manjaro" feel. as in being independent distribution).

If user doesn't need 1 file, they can just remove the packages. e.g `sudo pacman -R cachyos-kde-settings cachyos-alacritty-config`. It has zero effect on the system, as those config files get "used" only at user account creation(during CachyOS installation)

TheWyn:
>  It has zero effect on the system, as those config files get "used" only at user account creation(during CachyOS installation)

Removing `cachyos-kde-settings` turns off HDR for Sddm, that's a no go.  I'll fall back to suggest making it optional or removing.

PR #491, opened 2025-01-24
Bundle wine-mono and wine-gecko with wine-cachyos/wine-cachyos-opt

loathingKernel:
- wine-cachyos: bundle wine-gecko and wine-mono
- wine-cachyos-opt: bundle wine-gecko and wine-mono
- cachyos-gaming-meta: remove wine-gecko and wine-mono, bundled with wine-cachyos
- cachyos-gaming-meta: replace wine-cachyos with wine-cachyos-opt

### DIFF
--- a/cachyos-gaming-meta/.SRCINFO
+++ b/cachyos-gaming-meta/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = cachyos-gaming-meta
 	pkgdesc = Meta package for Gaming dependencies
 	pkgver = 1.0.0
-	pkgrel = 1
+	pkgrel = 2
 	epoch = 1
 	arch = any
 	depends = alsa-plugins
@@ -27,9 +27,7 @@ pkgbase = cachyos-gaming-meta
 	depends = proton-cachyos
 	depends = protontricks
 	depends = ttf-liberation
-	depends = wine-cachyos
-	depends = wine-gecko
-	depends = wine-mono
+	depends = wine-cachyos-opt
 	depends = winetricks
 	depends = vulkan-tools
 
--- a/cachyos-gaming-meta/PKGBUILD
+++ b/cachyos-gaming-meta/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=cachyos-gaming-meta
 pkgver=1.0.0
-pkgrel=1
+pkgrel=2
 epoch=1
 arch=(any)
 pkgdesc='Meta package for Gaming dependencies'
@@ -31,9 +31,7 @@ depends=(
   proton-cachyos
   protontricks
   ttf-liberation
-  wine-cachyos
-  wine-gecko
-  wine-mono
+  wine-cachyos-opt
   winetricks
   vulkan-tools
 )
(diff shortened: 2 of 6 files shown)


cachyos-gaming-meta/.SRCINFO
```diff
@@ -1,7 +1,7 @@
 pkgbase = cachyos-gaming-meta
 	pkgdesc = Meta package for Gaming dependencies
 	pkgver = 1.0.0
-	pkgrel = 1
+	pkgrel = 3
```
1Naim, line 4:
```suggestion
	pkgrel = 2
```

cachyos-gaming-meta/PKGBUILD
```diff
@@ -3,7 +3,7 @@
 
 pkgname=cachyos-gaming-meta
 pkgver=1.0.0
-pkgrel=1
+pkgrel=3
```
1Naim, line 6: same

1Naim (approved):
Aside from the skipped pkgrel, LGTM. I think it makes sense that a meta package should use a package that doesn't conflict with another system package (wine). Though I believe to make the transition a bit smoother we should provide a guide in the wiki or so on how to use the wine installed in /opt.

loathingKernel:
> Though I believe to make the transition a bit smoother we should provide a guide in the wiki or so on how to use the wine installed in /opt.

I agree, although I might not be the right person to do it because I do not daily-drive the various launchers people use. I can update the wiki generic instructions though to serve as a base example.
I do have plans to include an extra wrapper binary to be placed in `/usr/bin` to invoke wine from `/opt`.

Though I explained the `pkgrel` skip I am not particular to it, and if needed for clarity or anything, I can update them to not skip.

cachyos-gaming-meta/PKGBUILD
loathingKernel: This is due to two atomic commits, one to remove `wine-gecko` and `wine-mono`, and the second one to replace `wine-cachyos` with `wine-cachyos-opt`

cachyos-gaming-meta/.SRCINFO
loathingKernel: When I iterate over packages locally I bump the `pkgrel` to trigger my CI to build them. This is purely to be equal between what I track locally and what is in the Cachyos repos.

cachyos-gaming-meta/PKGBUILD
1Naim: Since both changes are in the same PR, please only bump the pkgrel by one increment. Skipping pkgrels can cause confusion to users and fellow maintainers.

1Naim:
> I can update the wiki with generic instructions though to serve as a base example.

Yes, that's perfectly fine.

1Naim:
I think this branch has some offtopic commits now. You should branch these off to have them separate from your work branch.

crian:
What is the status of this one? When will it be merged?

1Naim:
Right, are there any blockers for this (except the merge conflict of course, but that's trivial)? I would like to get this merged so we have parity with the wiki.

loathingKernel:
I have resolved the conflicts. I do not think there is anything outstanding blocking this right now.

ptr1337 (approved):

PR #494, opened 2025-01-26
cachyos-settings: Do not explicitly enable systemd-resolved after pac…

1Naim:
…kage update

It is sufficient to only enable systemd-resolved during first installs. Force enabling the service every update can be very annoying for users who explicitly do not want to use systemd-resolved for their DNS resolver.


vnepogodin (approved):
legit

ptr1337 (approved):

PR #510, opened 2025-02-11
kwin: 6.3.0

1Naim:


kwin/0001-libei-no-prompt.patch
```diff
@@ -592,134 +203,256 @@ index 84614eb51fa..5b27ec9b797 100644
...
++
  EisBackend::EisBackend(QObject *parent)
      : KWin::InputBackend(parent)
      , m_serviceWatcher(new QDBusServiceWatcher(this))
--
  {
- #if HAVE_XWAYLAND_ENABLE_EI_PORTAL
-     if (kwinApp()->operationMode() == Application::OperationModeXwayland) {
++#if HAVE_XWAYLAND_ENABLE_EI_PORTAL
++    if (kwinApp()->operationMode() == Application::OperationModeWayland) {
```
1Naim, line 345: So `OperationModeXwayland` was removed in 6.3, and I don't really know what the best alternative to this check is, so I opted to just use `OperationModeWayland` because you're most likely using xwayland anyway, and there's #if with xwayland

PR #534, opened 2025-02-27
feat: Split nvidia-open-dkms into its own package

1Naim:
Historically, NVIDIA's open kernel module GitHub repository has been lagging behind its runfile driver.
This has happened already for atleast 2 versions now, and it's uncertain whether or not their release model
will improve for future releases.|

This PR splits the open kernel module sourced from GitHub into its own package so it doesn't block users and/or packagers from updating the userspace drivers. Note that `nvidia-utils` now also supports packaging the open modules sourced from the runfile, although it is disabled by default to prefer the Github-sourced modules during packaging.


ptr1337:
https://download.nvidia.com/XFree86/NVIDIA-kernel-module-source/

We could also consider to use this, since there is a available checksum.

PR #535, opened 2025-02-28
nvidia-utils: Source open kernel modules from NVIDIA tarball

1Naim:
This tarball was brought to my attention by Peter in https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/534. At first glance, these tarballs are pushed
several days before NVIDIA makes their driver release announcement, and far earlier than what is pushed to
GitHub. On one hand, these sources are preferred over the GitHub sources because they are released earlier. On the other, some might prefer GitHub sources because of the code transparency. Diffs are great, but delayed releases are not, and it seems we can only choose one or the other.

Checksums are switched from sha512 to sha256 because the checksums provided by NVIDIA are sha256. No
functional changes intended.

Regardless of if we choose to switch to an alternative tarball source or not, I'd like to get the other three patches merged because IMO they seem to improve the PKGBUILD, especially https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/535/commits/1b74979421adca0c5210022568382663dbad89a7

Alternative to #534.


ryzendew (approved):
i approve of this 100% if it makes the nvidia experience better for the devs in anyway and stream lines the experience due to GH not always having the modules when the driver is out and yet they are here https://download.nvidia.com/XFree86/NVIDIA-kernel-module-source/.

ptr1337 (approved):

PR #543, opened 2025-03-09
loathing wine wow64 ntsync

loathingKernel:
- **wine: add pkguild**
- **wine-staging: add pkguild**
- **wine: build as wow64 wine, add ntsync patch**
- **wine-staging: build as wow64 wine, add ntsync patch**
- **linux-api-header: provide the ntsync-header name**


1Naim (commented):
I don't think we need to provide the pgpkeys here.

toolchain/linux-api-headers/PKGBUILD
```diff
@@ -21,6 +21,7 @@ sha256sums=('e79dcc6eb86695c6babfb07c2861912b635d5075c6cd1cd0567d1ea155f80d6e'
             'a208b8303d23d9b46e561a1be09498067e10f965c5d7814fac4c3b88c07e034f')
 validpgpkeys=('ABAF11C65A2970B130ABE3C479BE3E4300411886'   # Linus Torvalds
               '647F28654894E3BD457199BE38DBBDC86092693E')  # Greg Kroah-Hartman
+provides=('ntsync-header')
```
1Naim, line 24: Considering that this package should be obsolete by 6.14, should we also add replaces() here?

wine-staging/PKGBUILD
```diff
@@ -0,0 +1,172 @@
...
+_winever=$_pkgbasever
+#_winever=${_pkgbasever%.*}
+
+source=("git+https://gitlab.winehq.org/wine/wine.git?signed#tag=wine-$_pkgbasever"
+        "git+https://gitlab.winehq.org/wine/wine-staging.git#tag=v$_pkgbasever"
+        30-win32-aliases.conf
+        wine-binfmt.conf)
+source+=(
+        wine-ntsync.patch
+)
```
1Naim, line 22: I'm confused... why not just
```suggestion
source=("git+https://gitlab.winehq.org/wine/wine.git?signed#tag=wine-$_pkgbasever"
        "git+https://gitlab.winehq.org/wine/wine-staging.git#tag=v$_pkgbasever"
        30-win32-aliases.conf
        wine-binfmt.conf
        wine-ntsync.patch)
```

same with the rest of the arrays..

1Naim (commented):
Posted some reviews.... another thing I'm curious about is that do we need to include wine as well? I think most if not all users would just use -staging

toolchain/linux-api-headers/PKGBUILD
ptr1337: Im unsure, if we even need this provides? api-headers are installed on every system anyways

1Naim: Yeah, you're right. I think only `replaces()` is needed really, because `ntsync-headers` doesn't actually conflict with normal headers, and that it installs the headers to the correct path via pacman hooks.

loathingKernel: The PKGBUILD will ask for `ntsync-header`, which is provided by us through this package. If we accept that this PKGBUILD won't be used anywhere but CachyOS, we are safe to remove the `provides` here and the dependency from the wine PKGBUILDs. I don't think we should add replaces either, as the ntsync package was only in the AUR, cachyos never provided, correct?

wine-staging/PKGBUILD
loathingKernel: It helps with diffs as it creates less conflicts and more clear boundaries of the changes over the base. I also allows me to not change "unrelated" lines. The other reason is that these will likely be removed in a few versions when the patch is merged into mainline wine.

wine-staging/PKGBUILD
```diff
@@ -0,0 +1,172 @@
...
+  # Get rid of old build dirs
+  rm -rf $pkgname-{32,64}-build
+  mkdir $pkgname-{32,64}-build
+
+  cd wine
+  # apply wine-staging patchset
+  ../wine-staging/staging/patchinstall.py --backend=git-apply --all \
+    --no-autoconf \
+    -W eventfd_synchronization \
+    -W server-PeekMessage \
```
ventureoo, line 119: From my experience, `server-PeekMessage` is fine and it should not be excluded when ntsync is used.  But I could be wrong, maybe you had a different motivation for that.

wine-staging/PKGBUILD
```diff
@@ -0,0 +1,172 @@
...
+
+  echo "Building Wine-64..."
+  cd "$srcdir/$pkgname-64-build"
+  ../wine/configure \
+    --prefix=/usr \
+    --libdir=/usr/lib \
+    --with-x \
+    --with-wayland \
+    --with-gstreamer \
+    --with-mingw \
```
ventureoo, line 148: Not sure if we need to specify ``--with-mingw`` here since mingw is used by default.

wine/PKGBUILD
```diff
@@ -0,0 +1,157 @@
...
+  make
+}
+
+package() {
+  echo "Packaging Wine-64..."
+  cd "$srcdir/$pkgname-64-build"
+  make prefix="$pkgdir/usr" \
+    libdir="$pkgdir/usr/lib" \
+    dlldir="$pkgdir/usr/lib/wine" install
+
```
ventureoo, line 149: There's no strip. Maybe it was intended, but I prefer to clarify.

wine/PKGBUILD
```diff
@@ -0,0 +1,157 @@
...
+)
+optdepends+=(
+  NTSYNC-MODULE
+)
+makedepends=(${makedepends[@]} ${depends[@]})
+install=wine.install
+
+prepare() {
+  # Get rid of old build dirs
+  rm -rf $pkgname-{32,64}-build
```
ventureoo, line 107: Splitting into 32/64 bit is not needed here anymore, I guess.

wine-staging/PKGBUILD
```diff
@@ -0,0 +1,172 @@
...
+  export CROSSCFLAGS="-O2 -pipe"
+  export CROSSCXXFLAGS="-O2 -pipe"
+  export CROSSLDFLAGS="-Wl,-O1"
+
+  echo "Building Wine-64..."
+  cd "$srcdir/$pkgname-64-build"
+  ../wine/configure \
+    --prefix=/usr \
+    --libdir=/usr/lib \
+    --with-x \
```
ventureoo, line 145: I'm not sure if ffmpeg is actually used by default, so perhaps we should add ``--with-ffmpeg``here?

ventureoo (commented):
Just a few remarks from me. Nothing critical, though.

wine-staging/PKGBUILD
loathingKernel: Good point, so I checked, according to `configure` it should at least checked and found
```
checking for libavutil/avutil.h... yes
checking for libavformat/avformat.h... yes
checking for libavcodec/avcodec.h... yes
checking for libavcodec/bsf.h... yes
checking for av_log_set_callback in -lavutil... yes
checking for av_find_input_format in -lavformat... yes
checking for avcodec_get_name in -lavcodec... yes
```

wine/PKGBUILD
loathingKernel: Indeed

wine/PKGBUILD
loathingKernel: It is following the Arch PKGBUILD in what to do with debugging symbols. For reference: https://gitlab.archlinux.org/archlinux/packaging/packages/wine/-/merge_requests/2/diffs

wine-staging/PKGBUILD
loathingKernel: It is not required, but I had it in my own PKGBUILD. Will remove

wine/PKGBUILD
ptr1337: Can we not just disable it? I dont think we want to have a 600MB bigger package, till we have PE strip from pacman

wine-staging/PKGBUILD
loathingKernel: For me it fails to apply the ntsync patch, that's why I kept it out.

wine-staging/PKGBUILD
1Naim: Okay then, it looked insane to me but that's a fair argument. These packages are also maintained by you so whatever is most comfortable for you is what's best :)

wine-staging/PKGBUILD
ventureoo: All right, seems to be I just used to rebase ntsync over staging patchset, so it was not a problem for me. Let's just leave it at that.

wine-staging/PKGBUILD
ventureoo: Then it's okay, I guess.

toolchain/linux-api-headers/PKGBUILD
loathingKernel: Since we had it in the repos, then yes, using `replaces()` is applicable here. My question is about `provides()` mostly here, because if someone takes our PKGBUILD to build on their own on Arch, it will need the `ntsync-header` package, but not on CachyOS, (part of `base` as dependency of `glibc`) So `replaces()` should suffice here?

wine/PKGBUILD
ventureoo: > Can we not just disable it? I dont think we want to have a 600MB bigger package, till we have PE strip from pacman

Well, right way to go here is to use ``pestrip`` [1], but it's still not pushed to pacman package, so we're gonna be a bit stuck with that for a while.

- https://gitlab.archlinux.org/pacman/libmakepkg-dropins/-/blob/main/tidy/pestrip.sh

loathingKernel: > Can we not just disable it? I dont think we want to have a 600MB bigger package, till we have PE strip from pacman

The point of the MR I linked is to have a package that can allow users to report breakage to upstream Wine. Yeah, might be adding ntsync on top, but we add what is MR'd to wine, in an unmodified state too. So I think our Wine/Wine-staging should at this point in time facilitate a good base for issues to be reported to the upstream projects.

toolchain/linux-api-headers/PKGBUILD
loathingKernel: Removed any mention of `ntsync-header` from the affected packages as it is irrelevant to CachyOS.

1Naim (approved):
LGTM

ventureoo (approved):
LGTM, I think we can overlook on strip.

wine-staging/PKGBUILD
```diff
@@ -0,0 +1,168 @@
+# Maintainer: Felix Yan <felixonmars@archlinux.org>
+# Maintainer: Peter Jung <ptr1337@archlinux.org>
+# Contributor: Sven-Hendrik Haase <sh@lutzhaase.com>
+# Contributor: Jan "heftig" Steffens <jan.steffens@gmail.com>
+# Contributor: Eduardo Romero <eduardo@archlinux.org>
+# Contributor: Giovanni Scafora <giovanni@archlinux.org>
+
+pkgname=wine-staging
+pkgver=10.4
+pkgrel=1
```
1Naim, line 10: Can you set the pkgrel to 2, or +1 of Arch's version? Thanks.

PR #552, opened 2025-03-24
cachy-browser: 136.0.2

Stebalien:


Stebalien:
Note: I've updated the URLs/hashes/SRCINFO, but I haven't actually built it (my machine doesn't have enough ram for this beast). But the patches appear to apply.

ptr1337 (approved):

PR #576, opened 2025-04-14
feat: Rewrite package version check job

### LINKED ISSUE #396: Rewrite "update" bot
Adjusting [nvchecker-poc](https://gitlab.archlinux.org/archlinux/nvchecker-poc) for github API, adding exclude-list instead of include as in nvchecker-poc is done or doing own script should be sufficient

Current one https://github.com/CachyOS/CachyOS-PKGBUILDS/blob/master/.github%2Fworkflows%2Fmain.yml#L29

1Naim:
This new rewrite not uses the checker script written by @vnepogodin. It uses nvchecker to check against the upstream version and create a new issue if a package is outdated. The script also supports automatically closing issues when a version is already up-to-date.

Before merging this, we should validate all PKGBUILDs we have and remove `.nvchecker.toml` files that are not needed.

Closes #396


.github/scripts/checker.sh
```diff
@@ -0,0 +1,162 @@
...
+
+# --- Main execution ---
+pkgbuilds_dir="$1"
+if [[ -z "${pkgbuilds_dir}" ]]; then
+  echo "Usage: $0 <pkgbuilds_directory>"
+  exit 1
+fi
+
+if [[ "${pkgbuilds_dir}" == "--ci" ]]; then
+    files=$(find . -name "PKGBUILD")
```
vnepogodin, line 154: checking for .nvchecker.toml makes more sense here

vnepogodin (changes requested):

.github/scripts/checker.sh
1Naim: Right... goofy ahh

vnepogodin (approved):

vnepogodin:
yes.

PR #577, opened 2025-04-14
various: Add nvchecker

1Naim:
Adds nvchecker to various packages in this repo so the script in from #576 can cover it too.

P.S. please check if the contents of nvchecker is correct.


1Naim:
Another note: I haven't covered the CachyOS packages that are missing nvchecker. Since this can be considered low priority, I'll add them at a later time.

llvm/.nvchecker.toml
```diff
@@ -0,0 +1,3 @@
+[llvm]
+source = "archpkg"
+archpkg = "llvm"
```
1Naim, line 3: I may or may not have just realized this, but such checks against arch packages will be quite useless for rebuild notifications, since we already have pkgrel+1. Any thoughts of making the pkgrel `pkgrel.1` instead like the autobuilder?

vnepogodin: using only pkgver is the wanted solution here

1Naim: So we only want to check against actual version updates and not rebuilds? If so, I'll come back to this and add `strip_releases = true` for the toml files that are missing this.

vnepogodin: yes, for custom packages against archlinux repo

PR #591, opened 2025-04-22
[Cachy-Browser] Missing default privacy setting

TheWyn:
Turn off the chatbot by default to go with other privacy settings

Other settings checked already off by default:

- app.normandy.enabled
- app.normandy.api_url
- captivedetect.canonicalURL
- browser.ml.enable


TheWyn:
Yes you're right missed that repo will add there instead

TheWyn:
Moved to https://github.com/CachyOS/CachyOS-Browser-Settings/pull/25

PR #594, opened 2025-04-26
cachyos-hooks: 2025.04.26

ventureoo:
I don't know how I didn't notice it before, but it's a hotfix for script name in the hook.


vnepogodin (approved):
Fatal error

PR #632, opened 2025-05-15
proton-cachyos: fix crash in underrail

loathingKernel:


proton-cachyos/PKGBUILD
```diff
@@ -8,7 +8,7 @@ pkgver=${_srctag//-/.}
 _geckover=2.47.4
 _monover=10.0.0
 _xaliaver=0.4.6
-pkgrel=4
+pkgrel=5
```
ptr1337, line 11: Can you push to 6? Ive yesterday pushed to 5 already, since you wanted to have a rebuild with only default flags applied

loathingKernel: Το γιορύρ τς' Άρτας

https://en.wikipedia.org/wiki/Bridge_of_Arta#Folklore

loathingKernel: done

ptr1337 (approved):

PR #636, opened 2025-05-16
wine/proton 10.0-20250515

loathingKernel:
- **proton-cachyos: fix crash in underrail**
- **proton-cachyos: bump pkgrel**
- **proton-cachyos: version 10.0-20250515**
- **wine-cachyos: version 10.0-20250515, build as wow64**
- **wine-cachyos-opt: version 10.0-20250515, build as wow64**


ptr1337 (approved):
Lets see what breaks, thank you!

PR #638, opened 2025-05-17
added patch HW accelerated video playback causes VCN timeout on VCN 4…

romanstingler:
added patch from 
https://gitlab.freedesktop.org/mesa/mesa/-/issues/12528#note_2906598


1Naim:
I don't know if you're aware, but that patch you just added is for the kernel, not mesa. Regardless, I'v cherry-picked it https://github.com/CachyOS/linux/commit/b243d5dd250bfb4a9de96446cdc0d39d72cbd767. Thanks for the patch.

PR #651, opened 2025-05-22
cachyos-rate-mirrors: Check CDN77 availability by timeout

ventureoo:
It's possible to bypass Russian censorship by using tools such as zapret, which do not change the user's IP address and thus allow access to CDN77, but cachyos-rate-mirrors still will not use it.


vnepogodin (approved):
fair

PR #653, opened 2025-05-23
Do not strip for sign of extensions

oech3:
`Extra/code` stopped stripping.
https://gitlab.archlinux.org/archlinux/packaging/packages/code/-/issues/10

### DIFF
--- a/void/PKGBUILD
+++ b/void/PKGBUILD
@@ -9,7 +9,8 @@ pkgdesc="The open-source Cursor alternative"
 url="https://voideditor.com/"
 arch=('x86_64')
 license=("MIT")
-options=(!debug) # probably will be quite big
+options=(!strip # for sign of extension ?
+!debug) # probably big
 _elnum=34 # 35+ will fail
 depends=( electron${_elnum} ripgrep xdg-utils # replacements
   libx11


void/PKGBUILD
```diff
@@ -9,7 +9,8 @@ pkgdesc="The open-source Cursor alternative"
 url="https://voideditor.com/"
 arch=('x86_64')
 license=("MIT")
-options=(!debug) # probably will be quite big
+options=(!strip # for sign of extension ?
+!debug) # probably big
```
1Naim, line 13: ???????? what the fuck?
```suggestion
options=('!strip' '!debug')
```

oech3: How about comment? Some one may tries to remove them again.

loathingKernel: Comment on top properly then, with an explanation, not in-between and with cryptic one-liners. These comments mean nothing by themselves to any third party.

1Naim: who? or why? "probably big" isn't a good comment. I doubt anyone will ever touch pacman options here, and the commit history will speak for itself on why this was added in the first place, so your concern is unfounded.

PR #673, opened 2025-06-02
Move _gallium_xa to legacy following deprecation and frontend deletion

hivehivemind:
Gallium XA is deprecated and now disabled by default and will be removed in Mesa 25.2 https://gitlab.freedesktop.org/mesa/mesa/-/commit/31cf6b94ad1dfaf4272b22a39d7e2805d03f9375


ventureoo:
I think it can be simply removed since ``_gallium_xa`` doesn't have any effect after https://github.com/CachyOS/CachyOS-PKGBUILDS/commit/2c8b4c91173159b4f922763c30d1d27c7475c6e8 anyway.

PR #677, opened 2025-06-06
New package: cachyos-firefox-settings

ventureoo:
It's mostly based on my personal firefox settings, but doesn't necessarily need to be 100% duplicated, as I'm aware that some settings/policies can be re-enabled to improve usability.

### DIFF
--- a/cachyos-firefox-settings/PKGBUILD
+++ b/cachyos-firefox-settings/PKGBUILD
@@ -1,0 +1,29 @@
+# Maintainer: Vasiliy Stelmachenok <ventureo@cachyos.org>
+pkgname=cachyos-firefox-settings
+pkgver=20250607
+pkgrel=1
+pkgdesc='CachyOS settings for Firefox browser'
+depends=('firefox' 'xdg-desktop-portal-impl')
+arch=('any')
+backup=('etc/firefox/policies/policies.json')
+url="https://github.com/CachyOS/CachyOS-PKGBUILDS/tree/master/$pkgname"
+conflicts=(firefox-settings)
+replaces=(firefox-settings)
+provides=(firefox-settings)
+source=(
+    cachyos.js
+    policies.json
+)
+sha256sums=(
+    '2e3483172ac071960f6912ade26fdd66f2a8fe28ba4d27c066d544a7a67d23af'
+    '1e4c43ef30b5952df184859dc97bcab5fc34dbcd977d4852851d45a236e2f45d'
+)
+
+pkgver() {
+    date +%Y%m%d
+}
+
+package() {
+    install -Dm644 cachyos.js "$pkgdir/usr/lib/firefox/browser/defaults/preferences/cachyos.js"
+    install -Dm644 policies.json "$pkgdir/etc/firefox/policies/policies.json"
+}
--- a/cachyos-firefox-settings/cachyos.js
+++ b/cachyos-firefox-settings/cachyos.js
@@ -1,0 +1,224 @@
+// Use LANG environment variable to choose locale
+pref("intl.locale.requested", "");
+
+// Use system-provided dictionaries
+pref("spellchecker.dictionary_path", "/usr/share/hunspell");
+
+// Disable default browser checking.
+pref("browser.shell.checkDefaultBrowser", false);
+
+// Don't disable extensions in the application directory
+pref("extensions.autoDisableScopes", 11);
+
+// Enable hardware acceleration
+pref("media.hardware-video-decoding.force-enabled", true);
+pref("media.webrtc.hw.h264.enabled", true);
+pref("media.gpu-process-decoder", true);
+
+// Performance & Graphics Tweaks
+pref("gfx.canvas.accelerated.cache-items", 32768);
+pref("gfx.canvas.accelerated.cache-size", 4096);
+pref("gfx.content.skia-font-cache-size", 80);
+pref("gfx.webrender.all", true);
+pref("gfx.webrender.precache-shaders", true);
+pref("gfx.webrender.program-binary-disk", true);
+pref("layers.gpu-process.enabled", true); // Use dedicated GPU process
+
+// JavaScript & Process Tweaks
+pref("dom.ipc.forkserver.enable", true); // Enable Linux fork server for new processes
+pref("javascript.options.baselinejit.threshold", 50); // Lower JIT threshold
+pref("javascript.options.ion.threshold", 500); // Lower Ion threshold
+
+// Memory & Cache Tweaks (Non-Disk)
+pref("image.cache.size", 10485760); // Increase decoded image cache size (10MB)
+pref("image.mem.decode_bytes_at_a_time", 65536); // Increase image decode chunk size
+pref("image.mem.shared.unmap.min_expiration_ms", 120000);
+pref("media.cache_readahead_limit", 7200); // Increase media readahead
+pref("media.cache_resume_threshold", 3600); // Increase media resume threshold
+pref("media.memory_cache_max_size", 1048576); // Increase media memory cache (1GB)
+pref("media.memory_caches_combined_limit_kb", 3145728); // Increase combined media caches limit (3GB)
+pref("network.buffer.cache.size", 65535); // Network buffer size
+pref("network.ssl_tokens_cache_capacity", 32768); // Increase SSL token cache
+
+// Remove Mozilla stuff
+pref("app.normandy.api_url", "", locked);
+pref("app.normandy.enabled", false, locked);
+pref("breakpad.reportURL", "", locked);
+pref("browser.contentanalysis.default_result", 0, locked);
+pref("browser.contentanalysis.enabled", false, locked);
+pref("browser.crashReports.unsubmittedCheck.autoSubmit2", false, locked);
+pref("browser.discovery.enabled", false, locked);
+pref("browser.ml.enable", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.aboutpreferences", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.adsfeed", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.discoverystreamfeed", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.places", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.recommendationprovider", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.system.topsites", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.system.topstories", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.topsites", false, locked);
+pref("browser.newtabpage.activity-stream.system.showSponsored", false, locked);
+pref("browser.newtabpage.activity-stream.telemetry.structuredIngestion.endpoint", 0, locked);
+pref("browser.newtabpage.activity-stream.telemetry.ut.event", false, locked);
+pref("browser.newtabpage.activity-stream.unifiedAds.adsFeed.enabled", false, locked);
+pref("browser.newtabpage.activity-stream.unifiedAds.adsFeed.spocs.enabled", false, locked);
+pref("browser.newtabpage.activity-stream.unifiedAds.adsFeed.tiles.enabled",	false, locked);
+pref("browser.newtabpage.activity-stream.unifiedAds.endpoint", "", locked);
+pref("browser.newtabpage.activity-stream.unifiedAds.spocs.enabled",	false, locked);
+pref("browser.newtabpage.activity-stream.unifiedAds.tiles.enabled", false, locked);
+pref("browser.places.interactions.enabled", false, locked);
+pref("browser.tabs.crashReporting.sendReport", false, locked);
+pref("browser.uitour.enabled", false, locked);
+pref("browser.urlbar.suggest.addons", false, locked);
+pref("browser.urlbar.suggest.quicksuggest.fakespot", false, locked);
+pref("browser.urlbar.suggest.quicksuggest.nonsponsored", false, locked);
+pref("browser.urlbar.suggest.quicksuggest.sponsored", false, locked);
+pref("browser.urlbar.suggest.quicksuggest.topsites", false, locked);
+pref("browser.urlbar.suggest.trending", false, locked);
+pref("captivedetect.canonicalURL", "", locked);
+pref("devtools.debugger.remote-enabled", false, locked);
+pref("dom.private-attribution.submission.enabled", false, locked);
+pref("dom.push.enabled", false, locked);
+pref("dom.security.unexpected_system_load_telemetry_enabled", false, locked);
+pref("extensions.abuseReport.enabled", false, locked);
+pref("extensions.blocklist.enabled", false, locked);
+pref("extensions.getAddons.showPane", false, locked);
+pref("extensions.htmlaboutaddons.recommendations.enabled", false, locked);
+pref("extensions.pocket.enabled", false, locked);
+pref("media.gmp-gmpopenh264.autoupdate", false, locked);
+
+// Reduces the CLRite base refresh period in the lack of a persistent
+// background push service
+pref("security.OCSP.enabled", 0);
+pref("security.pki.crlite_mode", 2, locked);
+pref("security.remote_settings.crlite_filters.enabled", true, locked);
+pref("services.settings.poll_interval", 300);
+
+// Remove telemetry
+pref("browser.contentblocking.report.lockwise.enabled", false, locked);
+pref("browser.contentblocking.report.monitor.enabled", false, locked);
+pref("browser.discovery.enabled", false, locked);
+pref("browser.newtabpage.activity-stream.feeds.telemetry", false, locked);
+pref("browser.newtabpage.activity-stream.showSponsored", false, locked);
+pref("browser.newtabpage.activity-stream.showSponsoredTopSites", false, locked);
+pref("browser.newtabpage.activity-stream.telemetry", false, locked);
+pref("browser.newtabpage.activity-stream.telemetry.ut.events", false, locked);
+pref("browser.ping-centre.telemetry", false, locked);
+pref("browser.search.serpEventTelemetryCategorization.enabled", false, locked);
+pref("browser.search.serpEventTelemetryCategorization.regionEnabled", false, locked);
+pref("browser.search.update", false, locked);
+pref("browser.shopping.experience2023.enabled", false, locked);
+pref("browser.startup.homepage_override.mstone", "ignore", locked);
+pref("browser.urlbar.addons.featureGate", false, locked);
+pref("browser.urlbar.fakespot.featureGate", false, locked);
+pref("browser.urlbar.mdn.featureGate", false, locked);
+pref("browser.urlbar.pocket.featureGate", false, locked);
+pref("browser.urlbar.weather.featureGate", false, locked);
+pref("browser.urlbar.yelp.featureGate", false, locked);
+pref("datareporting.healthreport.uploadEnabled", false, locked);
+pref("datareporting.policy.dataSubmissionEnabled", false, locked);
+pref("datareporting.usage.uploadEnabled", false, locked);
+pref("extensions.getAddons.cache.enabled", false, locked);
+pref("network.captive-portal-service.enabled", false, locked);
+pref("network.connectivity-service.enabled", false, locked);
+pref("network.trr.confirmation_telemetry_enabled", false, locked);
+pref("security.certerrors.recordEventTelemetry", false, locked);
+pref("toolkit.coverage.endpoint.base", "", locked);
+pref("toolkit.coverage.opt-out", true, locked);
+pref("toolkit.telemetry.archive.enabled", false, locked);
+pref("toolkit.telemetry.bhrPing.enabled", false, locked);
+pref("toolkit.telemetry.cachedClientID", "", locked);
+pref("toolkit.telemetry.cachedProfileGroupID", "", locked);
+pref("toolkit.telemetry.coverage.opt-out", true, locked);
+pref("toolkit.telemetry.enabled", false, locked);
+pref("toolkit.telemetry.firstShutdownPing.enabled", false, locked);
+pref("toolkit.telemetry.newProfilePing.enabled", false, locked);
+pref("toolkit.telemetry.server", "data:,", locked);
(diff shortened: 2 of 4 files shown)


firefox-settings/policies.json
```diff
@@ -0,0 +1,92 @@
...
+    "EnableTrackingProtection": {
+      "Value": true,
+      "Fingerprinting": true,
+      "Cryptomining": true,
+      "EmailTracking": true
+    },
+    "DNSOverHTTPS": {
+      "Enabled": true,
+      "ProviderURL": "https://dnsbunker.org/dns-query"
+    },
```
1Naim: I don't think this should be here since other people might want to use different DNS providers.

firefox-settings/policies.json
```diff
@@ -0,0 +1,92 @@
...
+          "adminSettings": {
+            "selectedFilterLists": [
+              "adguard-cookies",
+              "adguard-generic",
+              "adguard-other-annoyances",
+              "adguard-popup-overlays",
+              "adguard-social",
+              "adguard-spyware",
+              "adguard-spyware-url",
+              "adguard-widgets",
```
1Naim: Same with this, I don't think we should be enabling too many filters OOTB. Just keep it default, or enable the two from [arkenfox](https://github.com/arkenfox/user.js/wiki/4.1-Extensions#-recommended) + EasyList Cookies.

firefox-settings/policies.json
ventureoo: I don't see this as a concern because this server is not hardcoded and anyone can change it in browser settings. Also DOH provider will be Cloudflare by default anyway.

firefox-settings/policies.json
ventureoo: IDK why anyone would want to see this web garbage, but OK, looks legit.

firefox-settings/policies.json
1Naim: Sure, my understanding was that anything set via policy will be hardcoded. If it's not then that's fine.

SoongVilda:
I have something similar for librewolf. would it be possible to do this also for librewolf?

**Path:**
```bash
/home/$USER/.librewolf/
```

**Config:**
```bash
/home/$USER/.librewolf/librewolf.overrides.cfg
```

```
// Performance & Graphics Tweaks
pref("gfx.canvas.accelerated", true);
pref("gfx.canvas.accelerated.cache-items", 32768);
pref("gfx.canvas.accelerated.cache-size", 4096);
pref("gfx.content.skia-font-cache-size", 80);
pref("gfx.webrender.all", true); // Force Webrender GPU acceleration
pref("gfx.webrender.precache-shaders", true);
pref("gfx.webrender.program-binary-disk", true);
pref("layers.gpu-process.enabled", true); // Use dedicated GPU process

// Memory & Cache Tweaks (Non-Disk)
pref("image.mem.decode_bytes_at_a_time", 65536); // Increase image decode chunk size
pref("image.mem.shared.unmap.min_expiration_ms", 120000);
pref("image.cache.size", 10485760); // Increase decoded image cache size (10MB)
pref("media.memory_cache_max_size", 1048576); // Increase media memory cache (1GB)
pref("media.memory_caches_combined_limit_kb", 3145728); // Increase combined media caches limit (3GB)
pref("media.cache_readahead_limit", 7200); // Increase media readahead
pref("media.cache_resume_threshold", 3600); // Increase media resume threshold
pref("network.buffer.cache.size", 65535); // Network buffer size
pref("network.ssl_tokens_cache_capacity", 32768); // Increase SSL token cache

// JavaScript & Process Tweaks
pref("javascript.options.baselinejit.threshold", 50); // Lower JIT threshold
pref("javascript.options.ion.threshold", 500); // Lower Ion threshold
pref("dom.ipc.forkserver.enable", true); // Enable Linux fork server for new processes

// UI/Behavior Tweaks
pref("apz.overscroll.enabled", true); // Enable overscroll effect
pref("general.smoothScroll", true); // Enable smooth scrolling
pref("general.smoothScroll.msdPhysics.continuousMotionMaxDeltaMS", 12);
pref("general.smoothScroll.msdPhysics.enabled", true); // Use physics-based smooth scrolling
pref("general.smoothScroll.msdPhysics.motionBeginSpringConstant", 600);
pref("general.smoothScroll.msdPhysics.regularSpringConstant", 650);
pref("general.smoothScroll.msdPhysics.slowdownMinDeltaMS", 25);
pref("general.smoothScroll.msdPhysics.slowdownMinDeltaRatio", "2.0"); // Ensure float format if needed
pref("general.smoothScroll.msdPhysics.slowdownSpringConstant", 250);
pref("general.smoothScroll.currentVelocityWeighting", "1.0"); // Ensure float format if needed
pref("general.smoothScroll.stopDecelerationWeighting", "1.0"); // Ensure float format if needed
pref("mousewheel.default.delta_multiplier_y", 300); // Adjust mouse wheel scroll speed
pref("content.notify.interval", 100000); // Reduce frequency of certain notifications

// Feature Enablement
pref("layout.css.grid-template-masonry-value.enabled", true); // Enable CSS Masonry layout
```

ventureoo:
> I have something similar for librewolf. would it be possible to do this also for librewolf?

I've been added some of our settings. Some of them have raised the Speedometer score quite nicely (+3-4 score).

SoongVilda:
Glad to hear. I spend pretty good time to make my LibreWolf and (alternatively Firefox) smoother and faster. :)

firefox-settings/policies.json
TheWyn: Cachy browser has these enabled https://github.com/CachyOS/CachyOS-Browser-Settings/blob/master/ublock-config-8.txt

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+pref("network.http.max-connections", 1800); //https://kb.mozillazine.org/Network.http.max-connections
+pref("network.http.max-persistent-connections-per-server", 10); //https://kb.mozillazine.org/Network.http.max-persistent-connections-per-server
+pref("network.http.max-urgent-start-excessive-connections-per-host", 5); //Number of connections that we can open beyond the standard parallelism limit defined by max-persistent-connections-per-server/-proxy to handle urgent-start marked requests
+pref("network.http.pacing.requests.enabled", false); //Disable pacing requests
+
+// Privacy hardening
+pref("browser.contentblocking.category", "strict"); // https://blog.mozilla.org/security/2021/02/23/total-cookie-protection/
+pref("browser.download.autohideButton", false, locked); //do not hide download button automatically
+pref("browser.download.start_downloads_in_tmp_dir", true, locked); //start downloads in temp
+pref("browser.helperApps.deleteTempFileOnExit", true); // https://bugzilla.mozilla.org/show_bug.cgi?id=238789#c19
```
therealmate: Default

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
+// Use LANG environment variable to choose locale
+pref("intl.locale.requested", "");
+
+// Use system-provided dictionaries
+pref("spellchecker.dictionary_path", "/usr/share/hunspell");
+
+// Disable default browser checking.
+pref("browser.shell.checkDefaultBrowser", false);
```
therealmate, line 8: Also default

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+// Don't disable extensions in the application directory
+pref("extensions.autoDisableScopes", 11);
+
+// Enable hardware acceleration
+pref("media.hardware-video-decoding.force-enabled", true);
+pref("media.webrtc.hw.h264.enabled", true);
+pref("media.gpu-process-decoder", true);
+
+// Performance & Graphics Tweaks
+pref("gfx.canvas.accelerated", true);
```
therealmate: Default

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+pref("gfx.canvas.accelerated.cache-items", 32768);
+pref("gfx.canvas.accelerated.cache-size", 4096);
+pref("gfx.content.skia-font-cache-size", 80);
+pref("gfx.webrender.all", true);
+pref("gfx.webrender.precache-shaders", true);
+pref("gfx.webrender.program-binary-disk", true);
+pref("layers.gpu-process.enabled", true); // Use dedicated GPU process
+
+// JavaScript & Process Tweaks
+pref("dom.ipc.forkserver.enable", true); // Enable Linux fork server for new processes
```
therealmate, line 28: Should be enabled by default in FF141, just letting you know

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+pref("browser.download.autohideButton", false, locked); //do not hide download button automatically
+pref("browser.download.start_downloads_in_tmp_dir", true, locked); //start downloads in temp
+pref("browser.helperApps.deleteTempFileOnExit", true); // https://bugzilla.mozilla.org/show_bug.cgi?id=238789#c19
+pref("dom.security.https_only_mode", true, locked); // only allow https in all windows, including private browsing
+pref("network.predictor.enable-prefetch", false); // Disable speculative website loading
+pref("network.predictor.enabled", false); // Disable speculative website loading
+pref("network.prefetch-next", false); // https://developer.mozilla.org/en-US/docs/Glossary/Prefetch
+pref("privacy.query_stripping.strip_list", "__hsfp __hssc __hstc __s _hsenc _openstat dclid fbclid gbraid gclid hsCtaTracking igshid mc_eid ml_subscriber ml_subscriber_hash msclkid oft_c oft_ck oft_d oft_id oft_ids oft_k oft_lk oft_sk oly_anon_id oly_enc_id rb_clickid s_cid twclid vero_conv vero_id wbraid wickedid yclid"); // https://groups.google.com/a/mozilla.org/g/dev-platform/c/1vOSas0ptVQ?pli=1
```
therealmate, line 204: Doesn't uBO block these already?

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+pref("urlclassifier.features.socialtracking.skipURLs", "*.instagram.com, *.twitter.com, *.twimg.com"); // allow embedded social content
+pref("urlclassifier.trackingSkipURLs", "*.reddit.com, *.twitter.com, *.twimg.com"); // allow embedded social content
+
+// Desktop integration
+pref("widget.use-xdg-desktop-portal.file-picker", 1);
+
+// UI/Behavior Tweaks
+pref("apz.overscroll.enabled", true); // Enable overscroll effect
```
therealmate: Default

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+pref("general.smoothScroll", true); // Enable smooth scrolling
```
therealmate: Default

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
+// Use LANG environment variable to choose locale
+pref("intl.locale.requested", "");
+
+// Use system-provided dictionaries
+pref("spellchecker.dictionary_path", "/usr/share/hunspell");
```
therealmate, line 5: Default

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+pref("browser.safebrowsing.provider.mozilla.gethashURL", "", locked);
+pref("browser.safebrowsing.provider.mozilla.lastupdatetime", 0, locked);
+pref("browser.safebrowsing.provider.mozilla.lists", "", locked);
+pref("browser.safebrowsing.provider.mozilla.lists.base", "", locked);
+pref("browser.safebrowsing.provider.mozilla.lists.content", "", locked);
+pref("browser.safebrowsing.provider.mozilla.nextupdatetime", 0, locked);
+pref("browser.safebrowsing.provider.mozilla.pver", 0, locked);
+pref("browser.safebrowsing.provider.mozilla.reportURL", "", locked);
+pref("browser.safebrowsing.provider.mozilla.updateURL", "", locked);
+pref("browser.safebrowsing.reportPhishURL", "", locked);
```
therealmate, line 185: Surely not all of these are required

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+pref("dom.push.enabled", false, locked);
+pref("dom.security.unexpected_system_load_telemetry_enabled", false, locked);
+pref("extensions.abuseReport.enabled", false, locked);
+pref("extensions.blocklist.enabled", false, locked);
+pref("extensions.getAddons.showPane", false, locked);
+pref("extensions.htmlaboutaddons.recommendations.enabled", false, locked);
+pref("extensions.pocket.enabled", false, locked);
+pref("media.gmp-gmpopenh264.autoupdate", false, locked);
+pref("security.pki.crlite_mode", 0, locked);
+pref("security.remote_settings.crlite_filters.enabled", false, locked);
```
therealmate: I don't think disabling security mechanisms are recommended, unless we aim for zero external requests on FF startup. I also don't see OCSP being disabled here, which is worse than CRLite. (since OCSP makes 3rd party requests, not just Mozilla requests)

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+
+// Remove telemetry
+pref("browser.contentblocking.report.lockwise.enabled", false, locked);
+pref("browser.contentblocking.report.monitor.enabled", false, locked);
```
therealmate, line 99: I don't think these send any telemetry, only collect them locally

cachyos-firefox-settings/cachyos.js
```diff
@@ -0,0 +1,224 @@
...
+
+// Privacy hardening
+pref("browser.contentblocking.category", "strict"); // https://blog.mozilla.org/security/2021/02/23/total-cookie-protection/
+pref("browser.download.autohideButton", false, locked); //do not hide download button automatically
+pref("browser.download.start_downloads_in_tmp_dir", true, locked); //start downloads in temp
+pref("browser.helperApps.deleteTempFileOnExit", true); // https://bugzilla.mozilla.org/show_bug.cgi?id=238789#c19
+pref("dom.security.https_only_mode", true, locked); // only allow https in all windows, including private browsing
+pref("network.predictor.enable-prefetch", false); // Disable speculative website loading
+pref("network.predictor.enabled", false); // Disable speculative website loading
+pref("network.prefetch-next", false); // https://developer.mozilla.org/en-US/docs/Glossary/Prefetch
```
therealmate, line 203: Pretty sure uBO disables these if the "disable prefetching" setting is enabled (which is enabled by default)

cachyos-firefox-settings/PKGBUILD
```diff
@@ -0,0 +1,29 @@
+# Maintainer: Vasiliy Stelmachenok <ventureo@cachyos.org>
+pkgname=cachyos-firefox-settings
+pkgver=20250606
+pkgrel=1
+pkgdesc='CachyOS settings for Firefox browser'
+depends=('firefox' 'xdg-desktop-portal')
```
therealmate: I think we need xdg-desktop-portal-impl instead of just xdg-desktop-portal

cachyos-firefox-settings/cachyos.js
ventureoo: It's not a default. It's just that the Firefox package in Arch Linux has these settings supplied (https://gitlab.archlinux.org/archlinux/packaging/packages/firefox/-/blob/main/PKGBUILD?ref_type=heads#L216-229), but that doesn't mean other versions of Firefox will have them.

cachyos-firefox-settings/cachyos.js
ventureoo: Same as in previous comment. Enabled in Arch, not Firefox.

cachyos-firefox-settings/cachyos.js
ventureoo: I prefer to leave it as is.

cachyos-firefox-settings/PKGBUILD
ventureoo: Not all XDG portal implementations provide a file picker, such as xdg-desktop-portal-wlr. In this case, there is no guarantee that this is the correct dependency. See https://github.com/ventureoo/PKGBUILDs/pull/4#discussion_r2119076072. I'd prefer to use xdg-desktop-portal-gtk, but that's not entirely accurate either.

cachyos-firefox-settings/cachyos.js
SoongVilda: Agree, but no problem to active them already, right?

cachyos-firefox-settings/cachyos.js
ventureoo: I agree that it's probably unnecessary, but I prefer to trust protection built into the browser more than an extension that isn't omnipresent even though it comes by default here.

cachyos-firefox-settings/cachyos.js
ventureoo: I don't like it either. Sorry, but these settings are based on my personal ones, so I don't want anything being collected without my awareness anyway.

cachyos-firefox-settings/cachyos.js
ventureoo: This was disabled because it needs `dom.push.enabled`, which is already disabled above. Yes, my motivation was to avoid unnecessary connections to Mozilla servers.

therealmate: Most users don't want to downgrade their security just to avoid a few checks. This isn't telemetry, it's a security feature. And I still don't see OCSP being disabled, even though it leaks browsing activity to the certificate authorities.

cachyos-firefox-settings/cachyos.js
therealmate: At this point, we are just removing features, then we should disable history, cookies, cache, I mean, they all collect data locally. Most users actually want to have features

1Naim: Is this a new feature in Firefox? This is my first time seeing it, and I've never seen anyone mention it until now.

Regardless, this package is still in its early stages. Heck, it isn't even in repos yet. The current settings are fine IMO as a starting point, we can work on fine tuning it after people start using it and get some feedback.

> Most users actually want to have features

If "most" of these users actually want this, they can override it via user.js right? If not then the option of removing this package still exists :) Noone is forcing anyone to use this. Settings are meant to be opiniated and there's bound to be disagreements.

cachyos-firefox-settings/cachyos.js
1Naim: Yeah I agree with @therealmate here, can we keep this enabled? I also don't see the point of disabling `dom.push.enable` either because:
a) isn't it behind a prompt?
b) it requires a subscription, so resetting `dom.push.userAgentID` should also just disable this feature.

cachyos-firefox-settings/cachyos.js
1Naim: I would prefer to keep this in too.

cachyos-firefox-settings/PKGBUILD
```diff
@@ -0,0 +1,29 @@
+# Maintainer: Vasiliy Stelmachenok <ventureo@cachyos.org>
+pkgname=cachyos-firefox-settings
+pkgver=20250607
+pkgrel=1
+pkgdesc='CachyOS settings for Firefox browser'
+depends=('firefox' 'xdg-desktop-portal')
+arch=('any')
+backup=('etc/firefox/policies/policies.json')
+url="https://github.com/CachyOS-PKGBUILDs/$pkgname"
```
1Naim: Is this intentionally a dead link?

cachyos-firefox-settings/cachyos.js
therealmate: So I looked into these prefs, and these are basically toggles for the banners in about:protections. They don’t collect anything, they’re barely visible, and disabling them only adds maintainer burden. Also, the monitor one is disabled by default anyway. No real reason to touch these in my opinion.

Screenshot_2025-06-07_13-37-14

cachyos-firefox-settings/cachyos.js
ventureoo: Again. My concern here was not against CRLite, but against the Push services that are needed to keep the CRLite database up to date. This causes TCP connections to be constantly held in the background and to me this is a significant privacy concern, no matter what Mozilla guarantees gives.

> a) isn't it behind a prompt?

Behind the prompt are only site's own permission to send notifications. The notifications themselves are constantly monitored and sent by Firefox push service, part of which is also updating databases like CRLite, but other uses of this service are possible.

> Firefox maintains an active connection to a push service in order to receive push messages as long as it is open. The connection ends when Firefox is closed. We store a randomized identifier (User Agent IDentifier or UAID) on our server for your browser, along with a random client-generated identifier for each push subscription. When you have any subscriptions, the UAID is required to allow our push service to route incoming messages to Firefox. If you don't have any active push notification subscriptions, Firefox rotates the UAID on each new connection. 
> ...
> We store your IP address for 90 days as part of this service

(https://support.mozilla.org/en-US/kb/push-notifications-firefox)

To reach a compromise here, I considered https://bugzilla.mozilla.org/show_bug.cgi?id=1880888, and concluded that:

1) We don't need to enable Push services for CRLite to work, as there is a timer here that automatically updates them.

2) The default timer timeout is 24 hours. I reduced this to 5 minutes in order to keep the database relatively fresh, and not have constant connections to the push service.

It's the optimal solution for me, so I don't think I want to revisit that in any way.

cachyos-firefox-settings/cachyos.js
ventureoo: > disabling them only adds maintainer burden

There is no burden of maintenance here for me as a maintainer of these settings. So no, I'll leave it here.

cachyos-firefox-settings/PKGBUILD
ventureoo: I decided to settle on xdg-desktop-portal-impl, since a hard dependency to one implementation doesn't look right anyway

cachyos-firefox-settings/cachyos.js
ventureoo: I think, yes? But extra layer of protection in the browser itself doesn't seem to be a issue.

ptr1337 (approved):

ptr1337:
Ill merge for now, for further improvements please send PR

PR #684, opened 2025-06-08
cachyos hyprland changes

GitUser-IamALIVE:
removed make added utilities


ptr1337:
Closing since we do not support the hyprland settings anymore

PR #701, opened 2025-06-19
feat(mesa): add valve gamescope limiter patch for mesa and lib32-mesa

Sterophonick:
This PR adds a patch that imports two commits from SteamOS. The first commit sets `vk_x11_override_min_image_count = 4` for Forza Horizon 5, and the second contains the Gamescope frame limiter patch, to allow for better support for overriding frame limits. As an example, legacy builds of Minecraft will render at hundreds of frames per second but with these patches, a gamescope session will force the game to run at the set refresh rate unless set otherwise.

This patch is intended for use with Handheld edition, but it does not affect desktop use.

It is not applied to mesa-git, as it would possibly have to be rebased extremely often.


mesa/lib32-mesa/valve.patch
1Naim, line 1: Does the 32 bit version actually need any of these patches, or realistically are they going to be used?

Sterophonick: SteamOS has this patch on the 32-bit Mesa. And I also still play a bunch of 32-bit games, mostly for Windows

1Naim: Okay then.

mesa/lib32-mesa/valve.patch
```diff
@@ -0,0 +1,134 @@
...
+             <option name="radv_disable_depth_storage" value="true"/>
+         </application>
++
++        <application name="Forza Horizon 5" application_name_match="ForzaHorizon5.exe">
++            <option name="vk_x11_override_min_image_count" value="4" />
++        </application>
+     </device>
+ </driconf>
+-- 
+GitLab
```
1Naim: Why is this not merged upstream?

1Naim: I'm not too confident on the FH5 patch as it could very well be too device-specific, and can then regress other devices. Can you change it so that you only include the gamescope fps limiter patches? Please also bump the pkgrel. Thanks.

Sterophonick: Fixed.

Sterophonick:
bumping this, wondering if this can get merged

Sterophonick:
I was kinda thinking that a `mesa-handheld` package could be more ideal for this use case, in which case i could reinstate the FH5 patch and maybe fixup the RADV Invariant patch to work there as well.

Sterophonick:
I was kinda thinking that a `mesa-handheld` package could be more ideal for this use case, in which case i could reinstate the FH5 patch and maybe fixup the RADV Invariant patch to work there as well.

Sterophonick:
Bazzite imports the same patches as SteamOS and I personally had backported some game fixes into that version previously. I'm not entirely sure but I *think* their Mesa was the same across all versions, for better or worse.

PR #712, opened 2025-06-26
new package: steam-jupiter-stable

Sterophonick:
This is the packaging for Steam as it is done for SteamOS, can be viewed in its original form here: https://gitlab.com/evlaV/jupiter_steam-jupiter-stable-PKGBUILD. The only change that was made was in `70-steam-jupiter-input.rules`, which was adapted to be friendly to users that are not named `deck`.

This package purports to solve https://github.com/CachyOS/CachyOS-Handheld/issues/56.


1Naim:
Can this be solved in the `cachyos-handheld` package instead? I don't think adding yet another new package is a good solution.

Sterophonick:
How would we wanna go about that? All that needs to happen is that Steam always launches with the `-steamdeck` flag

Sterophonick:
SteamOS handles it with an alternate `steam.sh` script that always uses the flag. They have their own package, but i think a oneshot systemd service or a pacman hook would suffice pretty well

PR #724, opened 2025-07-01
updated Resolve to 20.0.1

psygreg:
Updated PKGBUILD for latest release.


psygreg:
Found a way to fix the aforementioned issues by wiping out all Resolve data from `~/.local`, `~/.config` and `~/.cache`.

ptr1337 (approved):

PR #758, opened 2025-07-18
Plasma login manager update

VolRencs:
I synchronized with a package that was recently added to the archlinux repositories and also updated it to the latest commit, I don't know if I did it right or if I should have done it at all.
I need to test if everything works correctly.

If there's anything that needs to be corrected or added, tell me.


ptr1337:
Thanks, looks good. as soon this lands in the stable arch repos we can likely remove it

PR #760, opened 2025-07-19
Add GitHub desktop

VolRencs:
A much needed package for people who need a GUI for github. Tested it, everything works fine.
But people who have kwallet or its analogs disabled will not be able to authorize through github.


ptr1337:
Can we not use the AUR package there? The AUR package is also maintained by an official archlinux maintainer.

Updating AUR packages would be easier for us then customized ones :)

VolRencs:
Yes, you can use the AUR package, except I don't fully understand why python-setuptools is used there.

ptr1337:
Added from AUR

PR #761, opened 2025-07-20
lsfg-vk: update PKGBUILD and .install to reflect upstream changes

Pahheb:
- Update `PKGBUILD` to the most recent commit
- Add new dependencies and submodules
- Fix `DCMAKE_SHARED_LINKER_FLAGS` overriding `makepkg.conf`'s default `LDFLAGS`
- Remove `LTO` from `CFLAGS` (as it breaks linkage in `liblsfg-vk.so`)
- Modify `lsfg-vk.install` to reflect upstream changes


ptr1337 (approved):
Thank you much

PR #764, opened 2025-07-21
lsfg-vk: add recursive submodules && update source commits

Pahheb:
The most recent version requires recursive submodules.
- Update to the most recent commit
- Add `git submodule update --init --filter=tree:0 --recursive`

### DIFF
--- a/lsfg-vk/.SRCINFO
+++ b/lsfg-vk/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = lsfg-vk
 	pkgdesc = Lossless Scaling Frame Generation on Linux via DXVK/Vulkan
 	pkgver = r199.e67fcd3
-	pkgrel = 1
+	pkgrel = 2
 	url = https://github.com/PancakeTAS/lsfg-vk
 	install = lsfg-vk.install
 	arch = x86_64
--- a/lsfg-vk/PKGBUILD
+++ b/lsfg-vk/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=lsfg-vk
 pkgver=r199.e67fcd3
-pkgrel=1
+pkgrel=2
 pkgdesc="Lossless Scaling Frame Generation on Linux via DXVK/Vulkan"
 arch=('x86_64')
 url="https://github.com/PancakeTAS/lsfg-vk"
--- a/lsfg-vk/lsfg-vk.install
+++ b/lsfg-vk/lsfg-vk.install
@@ -106,7 +106,7 @@ post_install() {
 
     echo "================================================================================================================================="
     echo "==  To enable lsfg-vk, you can either use the environment variables listed here:                                               =="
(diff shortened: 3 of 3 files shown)


ventureoo:
I wonder, why it's not possible to just do ``git submodule update --init --filter=tree:0 --recursive``?

Pahheb:
> > > I wonder, why it's not possible to just do `git submodule update --init --filter=tree:0 --recursive`?
> > 
> > https://wiki.archlinux.org/title/VCS_package_guidelines#Git_submodules
> 
> We have used it for a while in proton-cachyos and it was not considered as a problem. Also I don't see any explanation there on how it works better than recursive update, but okay, you can do it according to the guideline, it just looks a bit complicated.

It is definitely an annoying practice, especially when you have dozens of submodules. Makes it harder to maintain, and requires you to figure out the aliases and commits each repo expects.

Depending on what method is preferred, I can adjust the PR accordingly.

ptr1337:
Im fine if we use ` git submodule update --init --filter=tree:0 --recursive` :) As long we not directly upstream to archlinux it should be fine, since its massive work

Pahheb:
> Im fine if we use ` git submodule update --init --filter=tree:0 --recursive` :) As long we not directly upstream to archlinux it should be fine, since its massive work

Awesome! I'll adjust it now!

Pahheb:
I have force-pushed the adjusted version as to avoid clouding up the history. Let me know if anything has to be changed!

Pahheb:
> Is it intended that lsfg-vk gets automatically enabled due the default config? Or is this an upstream issue
> 
> ```
> cat conf.toml 
> version = 1
> [global]
> # override the location of Lossless Scaling
> # dll = "/games/Lossless Scaling/Lossless.dll"
> 
> # [[game]] # example entry
> # exe = "Game.exe"
> #
> # multiplier = 3
> # flow_scale = 0.7
> # performance_mode = true
> # hdr_mode = false
> #
> # experimental_present_mode = "fifo"
> 
> [[game]] # default vkcube entry
> exe = "vkcube"
> 
> multiplier = 4
> performance_mode = true
> 
> [[game]] # default benchmark entry
> exe = "benchmark"
> 
> multiplier = 4
> performance_mode = false
> 
> [[game]] # override Genshin Impact
> exe = "Genshin"
> 
> multiplier = 3
> ```

It is an upstream change, yes, and it's intended. 

I decided not to alter that functionality, as it could throw people off when consulting the wiki. 

Though, if it is frowned upon, there's the possibility of installing an empty config file if one does not exist yet, so new users don't encounter this behavior.

loathingKernel:
> Apparently it goes against AUR guidelines, as then the end user does not know explicitly what sources are being pulled in

Technically this is wrong. You can pull arbitrary sources in the way the Arch Wiki suggests, but by using recursive you solely depend on what the git history tells you to do.

I have had discussions about this in the wiki's discussion sections, and generally it wasn't frown upon as a solution when submodules get overwhelming.

PR #804, opened 2025-08-12
nvidia-utils: 580.65.06-1

ptr1337:
Signed-off-by: Peter Jung <admin@ptr1337.dev>


nvidia/nvidia-utils/gsk-renderer.sh
```diff
@@ -0,0 +1 @@
+export GSK_RENDERER=ngl
```
ventureoo, line 1: We should probably check if this is a hybrid graphics system: 
```bash
[[ -n "$(lspci -d "10de:*:0302")" ]] && export GSK_RENDERER=ngl
```
Otherwise this will lead to the use of OpenGL when using iGPU, which is probably not desirable.

ptr1337: Sure, we can do this. even tough it should not make any issues using for a few weeks the ngl backend

PR #821, opened 2025-08-24
Bump cachy-update to include latest patch commit & add missing sudo-rs optdepend

Antiz96:
- Latest patch commit fixing tooltip's branding: https://github.com/CachyOS/cachy-update/pull/3
- `sudo-rs` support added upstream in: https://github.com/Antiz96/arch-update/pull/408


ptr1337:
Thank you very much :pray:

PR #836, opened 2025-08-28
ci: update validate pkgbuilds

azdanov:
This should only run when a PR has PKGBUILD changes and only for the package that is being added in the PR. On push to the main branch, the previous validation of all packages should still happen. The same applies to manually triggering the workflow in the Actions tab.

cc @vnepogodin @SoulHarsh007 

Closes #835

### DIFF
--- a/.github/scripts/validate-pkgbuilds.js
+++ b/.github/scripts/validate-pkgbuilds.js
@@ -123,9 +123,14 @@ if (processedPkgbuilds.length > 0) {
     .sort((a, b) => a.pkgName.localeCompare(b.pkgName))
     .forEach((pkg) => {
       console.log(`PKGBUILD: ${pkg.pkgName}`);
-      pkg.details.forEach((detail) => {
-        console.log(`  [${detail.type}] ${detail.code}`);
-      });
+      pkg.details
+        .sort(
+          (a, b) =>
+            a.type.localeCompare(b.type) || a.code.localeCompare(b.code),
+        )
+        .forEach((detail) => {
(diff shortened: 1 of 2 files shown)


.github/scripts/validate-pkgbuilds.js
```diff
@@ -2,14 +2,55 @@ import { readdirSync } from "node:fs";
...
+  return stdout
+    .toString("utf8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((file) => file.endsWith("/PKGBUILD"))
+    .map((file) => join(pkgbuildsDir, file, ".."));
+}
+
+async function getAllPkgbuilds() {
```
SoulHarsh007: This should be a Synchronous function, there is no async or await being used here!

.github/scripts/validate-pkgbuilds.js
```diff
@@ -2,14 +2,55 @@ import { readdirSync } from "node:fs";
...
+    await $`git diff --name-only ${process.env.BASE_REF} ${process.env.HEAD_REF}`
+      .nothrow()
+      .quiet();
+
+  return stdout
+    .toString("utf8")
+    .trim()
+    .split("\n")
+    .filter(Boolean)
+    .filter((file) => file.endsWith("/PKGBUILD"))
```
ventureoo: Just FYI, it can be done a little easier by simply adding a pattern at the end of git command:
https://github.com/ventureoo/PKGBUILDs/blob/main/.github/workflows/repo.yml#L67

ventureoo (approved):

PR #840, opened 2025-08-30
Update PKGBUILD to include network-manager-applet

commonsourcecs:
Add network-manager-applet to depends so that autostart.conf works as intended.


ArjixWasTaken:
Since you are not modifying the version, you should bump the `pkgrel`.
You should also update the `.SRCINFO` (`$ makepkg -do --printsrcinfo`)

commonsourcecs:
You are quite right. 
Those things are now done.

PR #845, opened 2025-09-04
cachyos-gaming-meta: mark a few packages as explicit after installation

loathingKernel:


vnepogodin:
I'm really worried about that workaround

loathingKernel:
Yeah, it is not ideal, but what worries you exactly?

ventureoo:
> > Can't we actually just convert cachyos-gaming-meta into a package group?
> 
> In general this would be a good concept I think, but I don't know if it solve our orphan issue. We need to mark some packages as explicitly installed somehow.

All packages installed from package group are marked as explicitly installed except for their dependencies.

loathingKernel:
So where do we stand on this? Do we have any ideas for an alternative approach?

ventureoo:
> So where do we stand on this? Do we have any ideas for an alternative approach?

From my point of view, it’s better to merge it now to fix the current installations and leave cachyos-gaming-meta as is, and for new installations make a package group - cachyos-gaming.

ptr1337:
I think we could remove having proton-cachyos explicitly installed, since the "goal" would be defaulting to SLR. If people remove orphans, then this would bring the state we want.

ptr1337:
How should we proceed here now?

loathingKernel:
If this actually works, in effect, it does mark the packages as explicit, then it won't cause any issues. We can merge it and figure out how to improve it as a whole later.

loathingKernel:
Gentle bump

PR #862, opened 2025-09-20
bcachefs: use UUID for rootfs in kernel cmd line

jackwasey:
As of September 2025, CachyOS default install with a single-drive bcachefs filesystem for root works. However, with more than one disk in the bcachefs filesystem, sdboot-manage uses the colon-delimited list of devices for root= in the kernel command line, which fails. This patch will find the bcachefs filesystem UUID, and use that for root= instead.


systemd-boot-manager/sdboot-manage
1Naim, line 129: Are you implying that both of these are an empty string for your root filesystem?

jackwasey: -n is --noheadings, -o is --output

1Naim: Thanks but that doesn't answer my question. If you're getting the list of devices in `root=` instead of UUIDs, then it means that both those commands are an empty string for bcachefs or your particular setup. If its the former then I'd like to see this merged, but if its the latter then this solution is unsufficient since there needs to be a more generic way to solve this, or does this only happen on bcachefs, and not other filesystems?

jackwasey: Sorry, I misunderstood completely. Yes, I get empty string for root for both UUID and PARTUUID. This is run on a multi-device bcachefs root with machine booted using UUID for rootfs in command line (I can't boot any other way), so i think the reverse lookup from the multi-device string returned as SOURCE for findmnt -T / fails with the current findtools. I hope this answers your question.

systemd-boot-manager/sdboot-manage
```diff
@@ -146,6 +146,41 @@ generate_entries() {
         btrfs)
             sdoptions+=" rootflags=subvol=${srcdev_fsroot}"
             ;;
+        bcachefs)
+            bch_member="${srcdev%%:*}"
+            if [ -z "${bch_member}" ]; then
+                echo "Failed to get a bcachefs root filesystem member device"
```
ventureoo:
```suggestion
                die "Failed to get a bcachefs root filesystem member device"
```

systemd-boot-manager/sdboot-manage
```diff
@@ -146,6 +146,41 @@ generate_entries() {
...
+            if ! [ -b "${bch_member:-}" ]; then
+                echo "Failed to get a valid bcachefs root filesystem member device: not a block device."
+                return 1
+            fi
+            bch_uuid="$(
+                blkid --match-types=bcachefs --output export "${bch_member}" |
+                sed -ne '/^UUID=/s///p'
+            )"
+            if [ -z "${bch_uuid}" ]; then
+                echo "Failed to get bcachefs root filesystem UUID"
```
ventureoo:
```suggestion
                die "Failed to get bcachefs root filesystem UUID"
```

systemd-boot-manager/sdboot-manage
```diff
@@ -146,6 +146,41 @@ generate_entries() {
...
+            # list of bcachefs root filesystem members.
+            bch_check="$(
+                set +e +o pipefail
+                blkid --output export --uuid="${bch_uuid}" |
+                    sed -ne '/^UUID=/s///p'
+            )"
+            case "${srcdev}" in
+                *${bch_check}*) : ;;
+                *)
+                    echo "Failed to confirm correct UUID and member devices for bcachefs root filesystem."
```
ventureoo:
```suggestion
                    die "Failed to confirm correct UUID and member devices for bcachefs root filesystem."
```

PR #877, opened 2025-09-26
stacer: add package

VolRencs:
Prepared pkgbuild for stacer, tested it, it works for me.


ptr1337 (approved):

ptr1337:
Thank you! Will push tmrw

Would be also good to add nvchecker, but i can do this too

PR #883, opened 2025-09-29
ci: fix PR changes not detected by validate-pkgbuilds

azdanov:
At the moment, the PRs are always green because the BASE and HEAD refs are the same (`origin/master`), which doesn't detect any changes. E.g., https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/871/checks

image

This should hopefully fix the issue by using the commit's SHA instead of what was used before.

Example test run https://github.com/azdanov/CachyOS-PKGBUILDS/actions/runs/18100784454/job/51502912207


vnepogodin:
indeed. should fixed cases like [here](https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/882/checks)

PR #900, opened 2025-10-12
sdboot-manage: Add extra check before removing orphan entry

marioroy:
The "sdboot-manage remove" command may remove valid entries.

In the case having "linux-cachyos" and "linux-cachyos-evvdf" kernels, uninstalling "linux-cachyos" followed with "sdboot-manage remove" will also wipe the "linux-cachyos-eevdf" boot configuration file.

This is because the shorter kernel name matches the eevdf entry. Add check ensuring the "/usr/src/[kernel]" dir does not exists.


ventureoo:
Since we support this feature, I would still like to pull this fix.

vnepogodin (approved):
Many thanks!

PR #903, opened 2025-10-12
[zstd] Enable vectorized XXH64 for `znver4`

NexusXe:
As of facebook/zstd#3819, "auto vectorization of xxhash64" is disabled when AVX-512 is present due to the high latency of the Skylake-X implementation of AVX-512.

Other Intel uarchs also suffer from this high latency, but `znver4` and `znver5` do not, according to uops.info.

This PR re-enables auto vectorization for XXH64 on builds targeting `znver4` by conditionally adding the `XXH_ENABLE_AUTOVECTORIZE` flag according to the `CARCH` variable in the `PKGBUILD` script for `zstd`, `zstd-pgo`, and `lib32-zstd`.

According to [facebook/zstd/lib/common/xxhash.h](https://github.com/facebook/zstd/blob/dev/lib/common/xxhash.h), "Autovectorization of XXH64 tends to be detrimental, though the exact outcome may change depending on exact cpu and compiler version. For information, it has been reported as detrimental for Skylake-X, but possibly beneficial for Zen4."

### DIFF
--- a/zstd/lib32-zstd/PKGBUILD
+++ b/zstd/lib32-zstd/PKGBUILD
@@ -3,6 +3,8 @@
 # Contributor: Andrzej Giniewicz <gginiu@gmail.com>
 # Contributor: Johan Förberg <johan@forberg.se>
 
+_optimize_znver4=${_optimize_znver4-}
+
 _pkgname=zstd
 pkgname=lib32-zstd
 pkgver=1.5.7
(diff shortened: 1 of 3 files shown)


zstd/lib32-zstd/PKGBUILD
```diff
@@ -48,14 +48,25 @@ build() {
   export CFLAGS+=' -ffat-lto-objects'
   export CXXFLAGS+=' -ffat-lto-objects'
 
+  # https://github.com/facebook/zstd/issues/3819
+  # XXH64_update can be vectorized with AVX-512, but this is disabled by default
+  # due to Skylake's high latency for AVX-512 instructions. Re-enable this for
+  # Znver4, since unlike Intel u-archs, it doesn't suffer from high AVX-512 latency.
+  xxh64_vectorize=''
+
+  if  [[ "$CARCH" == "znver4" ]]; then
```
ptr1337: Thats tricky. There is currently no CARCH set, and the CARCH is overwritten in the packaging, since at archlinux a lot of stuff depends on have CARCH=x86_64. 

Maybe add a if statement, like we have here:
https://github.com/CachyOS/CachyOS-PKGBUILDS/blob/master/go/PKGBUILD#L18

which we then sed before the build for znver4

NexusXe: So, is `_optimize_znver4` set when building for Zen 4 repos? Or do I have to configure that to be set somewhere?

ptr1337: Yes

PR #921, opened 2025-10-17
bump: update ripgrep to version 15.0.0 and adjust dependencies

luadebug:
Resolves: https://github.com/CachyOS/CachyOS-PKGBUILDS/issues/916

```
~/CachyOS-PKGBUILDS/ripgrep mesa
❯ 
rg --version
ripgrep 15.0.0 (rev 87929f1150)

features:+pcre2
simd(compile):+SSE2,+SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,-AVX2

PCRE2 10.45 is available (JIT is available)
```

### DIFF
--- a/ripgrep/.SRCINFO
+++ b/ripgrep/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = ripgrep
 	pkgdesc = A search tool that combines the usability of ag with the raw speed of grep
 	pkgver = 15.0.0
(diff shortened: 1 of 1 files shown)


ptr1337:
Just a note for upcoming updates - for packages present in the archlinux repository, we need to choose a higher one. So we use archlinux pkgrel +1, in this case this would be then pkgrel=2

PR #1008, opened 2025-12-02
nvidia-utils: 590.44.01-1

ptr1337:
Not tested yet

Also the replacement needs to be better done, we might need to allign with archlinux pushing into the stable repository.

We need to find a way to get a proper migration for cachyos users too. A lot of depends each other.

New libary:

libnvidia-tileiras.so --> No idea what it does


nvidia/lib32-nvidia-utils/PKGBUILD
```diff
@@ -5,16 +5,16 @@
 _pkgbasename=nvidia-utils
 pkgbase=lib32-$_pkgbasename
 pkgname=('lib32-nvidia-utils' 'lib32-opencl-nvidia')
-pkgver=580.105.08
-pkgrel=2
+pkgver=590.44.01
+pkgrel=0
```
ventureoo, line 9: I think it's a typo?
```suggestion
pkgrel=1
```

nvidia/nvidia-utils/PKGBUILD
```diff
@@ -331,8 +314,9 @@ package_nvidia-open-dkms() {
   pkgdesc="NVIDIA open kernel modules - module sources"
   depends+=('dkms')
   license=('MIT AND GPL-2.0-only')
-  conflicts=('nvidia-open' 'NVIDIA-MODULE')
-  provides=('nvidia-open' 'NVIDIA-MODULE')
+  conflicts=('nvidia-open' 'NVIDIA-MODULE' 'nvidia-dkms')
+  provides=('nvidia-open' 'NVIDIA-MODULE' 'nvidia-dkms')
+  replaces=('nvidia-dkms')
```
ventureoo, line 319: `conflicts` and `provides` are fine, but I'm not sure if should specify `replaces` here if we want to make sure that for Pascal/Maxwell users nvidia-580xx-dkms will replace nvidia-dkms.

nvidia/nvidia-utils/PKGBUILD
```diff
@@ -248,6 +228,9 @@ package_nvidia-utils() {
     # Present Helper
     install -Dm755 "libnvidia-present.so.${pkgver}" "${pkgdir}/usr/lib/libnvidia-present.so.${pkgver}"
 
+    # Present Helper
+    install -Dm755 "libnvidia-tileiras.so.${pkgver}" "${pkgdir}/usr/lib/libnvidia-tileiras.so.${pkgver}"
```
ventureoo, line 232: I'm not aware what exactly it is, but in ``sandboxutils-filelist.json`` it was placed under `cuda` category.

nvidia/lib32-nvidia-utils/PKGBUILD
ventureoo, line 91: ``libnvidia-tileiras.so`` also has 32-bit version, should be added here.

ventureoo (changes requested):

nvidia/lib32-nvidia-utils/PKGBUILD
ptr1337: intended for testing

nvidia/nvidia-utils/PKGBUILD
1Naim: Don't need to explicitly conflict and provide `nvidia-dkms` because that provides `NVIDIA-MODULE`. Are we going to have the 580 drivers in repo? If yes then we should first migrate closed users to that package. Regardless of that, IMO we should be rid of the `-open` suffix, since theres only one variant now.

ptr1337: I did not think much about this yet, so this needs rework anyways

nvidia/lib32-nvidia-utils/PKGBUILD
ptr1337: 90mb has the binary, lot of AI bloat :(

ventureoo: Then split it into another package?

ptr1337: No no, we can just put it there. I just do not see any usecase yet tho, but we can package it

ptr1337:
Added nvidia-580xx. This was inspired by nvidia-550xx and followed the same scheme.

nvidia/nvidia-utils/PKGBUILD
ptr1337: > Don't need to explicitly conflict and provide `nvidia-dkms` because that provides `NVIDIA-MODULE`. Are we going to have the 580 drivers in repo? If yes then we should first migrate closed users to that package. Regardless of that, IMO we should be rid of the `-open` suffix, since theres only one variant now.

So, the nvidia-closed module is still provided by NVIDIA but the testing is bare minimum. This driver could be only used to disable GSP Firmware.

pallaswept:
> libnvidia-tileiras.so --> No idea what it does

Just guessing by the name, I presume that this is a lib for supporting their new CUDA Tile API:

https://developer.nvidia.com/blog/focus-on-your-algorithm-nvidia-cuda-tile-handles-the-hardware

Hope that helps.

PR #1018, opened 2025-12-05
nvidia-580xx: Introduce package

ventureoo:
Split out from https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/1008, `replaces` is specified to perform the migration for Pascal/Maxwell/Volta users.


ventureoo:
> Seems good to me with replacing all the kernels 👍
> 
> I think you forgot the default kernel "linux-cachyos" and "linux-cachyos-lto" in the replacement

Oops, yes, forgot to correct search pattern. Fixed.

ptr1337:
> LGTM. Are we going to keep the nvidia-open scheme for 590? Since that's the only variant left it doesn't really make sense to keep it and I don't think the migration would be too bad.

I think this could make the transition really really heavy, if we rename the 590 nvidia-open to nvidia.

ventureoo:
> LGTM. Are we going to keep the nvidia-open scheme for 590? Since that's the only variant left it doesn't really make sense to keep it and I don't think the migration would be too bad.

At the time of migration, I don’t think so. After a while - maybe.

I also want to note that I have made `lib32-nvidia-580xx-utils` a hard dependency for `nvidia-580xx-utils` right now. This may lead to issues on systems without multilib, but I don’t see any other way to automatically replace `lib32-nvidia-utils`.

PR #1034, opened 2025-12-18
Handheld: Replace SDDM with Plasma Login Manager

Revival8697:
See https://github.com/CachyOS/CachyOS-Handheld/pull/88.


1Naim (approved):

Revival8697:
You are supposed to review https://github.com/CachyOS/CachyOS-Handheld/pull/88 first though.

Someone should test this on a device other than the Steam Deck and see if:
- Session switching works, Plasma is killed after switching to Gamescope and vice versa.
- Autologin is not exposed in settings.

PR #1035, opened 2025-12-18
cachyos-calamares: Build package from unified branch

ventureoo:
Depends on: https://github.com/CachyOS/cachyos-calamares/pull/128


ptr1337:
Adding the replaces is not good, since we dont really want to replace anything, otherwise this would break the "older" ISO, which we generally keep for one more month after release.

This is anyways only handled by our scripts with installing anyways, so i think we can remove this

PR #1039, opened 2025-12-18
nvidia-utils: ship CUDA perf boost overrides

ptr1337:
- Install 50-nvidia-cuda-disable-perf-boost.conf via environment.d to set CUDA_DISABLE_PERF_BOOST=1 (reduce idle power with CUDA contexts like NVDEC/NVENC).
- Add cuda-no-stable-perf-limit application profile to allow full perf while streaming/recording (OBS/Discord/etc; NVIDIA/open-gpu-kernel-modules#333).


nvidia/nvidia-utils/cuda-no-stable-perf-limit
```diff
@@ -0,0 +1,24 @@
...
+            "settings": ["0x166c5e", 0]
+        }
+    ],
+    "rules": [
+        {
+            "pattern": {
+                "op": "or",
+                "sub": [
+                    { "feature": "procname", "matches": "obs" },
+                    { "feature": "procname", "matches": "Discord" },
```
1Naim, line 14: Does discord even support NVENC?

ptr1337: Should, yep

1Naim: I checked and it doesn't yet. But we can keep it until it eventually does I guess, doesn't seem too far into the future. It might not work on the custom clients though.

notpentadactyl:
> We may want to add more applications, like Discord forks and web browsers

Agree, should probably have the same ones as here: `/etc/nvidia/nvidia-application-profiles-rc.d/limit-vram-usage`

`firefox`, `brave`, `chrome`, `chromium`

nvidia/nvidia-utils/cuda-no-stable-perf-limit
```diff
@@ -0,0 +1,24 @@
+{
+    "profiles": [
+        {
+            "name": "CudaNoStablePerfLimit",
```
1Naim, line 4: should add ffmpeg here too.

nvidia/nvidia-utils/cuda-no-stable-perf-limit
```diff
@@ -0,0 +1,24 @@
...
+        {
+            "pattern": {
+                "op": "or",
+                "sub": [
+                    { "feature": "procname", "matches": "obs" },
+                    { "feature": "procname", "matches": "Discord" },
+                    { "feature": "procname", "matches": "discord" },
+                    { "feature": "procname", "matches": "DiscordCanary" },
+                    { "feature": "procname", "matches": "vesktop" },
+                    { "feature": "procname", "matches": "webcord" }
```
ventureoo, line 21: I think we should add here apps like `gpu-screen-recorder` and `simplescreenrecorder` too. In question are spectacle and built-in gnome screencast recorder, since I don't actually know if they're using NVENC.

ptr1337: done

PR #1051, opened 2025-12-29
Revert moving proton-cachyos to /opt

loathingKernel:
- **proton-cachyos: revert move to /opt**
- **proton-cachyo-slr: revert move to /opt**


DocMAX:
What stupid action was this???

PR #1130, opened 2026-02-12
cachyos-rate-mirrors: Check RU mirrors availability before inserting 

ventureoo:
Unfortunately, all mirrors in Russia are broken right now, so we should not force it.


ptr1337:
Cant we add a check if theyre available?

ptr1337:
> > Cant we add a check if theyre available?
> 
> This is a good idea and we can, only then it will require manually running cachyos-rate-mirrors after installing system if mirror was not available at the time.

Can we put in the if statement a curl check or ping check if its fine?

I think people still can rank cdn77 and clouldflare mirrors in russia, but when they try to pull it fails, if im correct

ptr1337 (approved):

PR #1144, opened 2026-02-17
cachyos-rate-mirrors: Add fallback in case geoip.kde.org is down

ptr1337:
geoip.kde.org was today down resulted into failed ranking of the mirrors as well as issues at installation. Add fallback

### DIFF
--- a/cachyos-rate-mirrors/cachyos-rate-mirrors
+++ b/cachyos-rate-mirrors/cachyos-rate-mirrors
@@ -104,7 +104,12 @@ rate_repository_mirrors() {
 
 # Prefer mirrors that are geographically close, especially in countries that
 # have regional blockages
-country="$(curl --connect-timeout 10 -sSL 'https://geoip.kde.org/v1/ubiquity' | grep -Po '<CountryCode>\K([A-Z]{2})')"
+country="$(curl --connect-timeout 10 -sSL 'https://geoip.kde.org/v1/ubiquity' 2>/dev/null | grep -Po '<CountryCode>\K([A-Z]{2})')"
+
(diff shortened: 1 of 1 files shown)


cachyos-rate-mirrors/cachyos-rate-mirrors
```diff
@@ -104,7 +104,12 @@ rate_repository_mirrors() {
 
 # Prefer mirrors that are geographically close, especially in countries that
 # have regional blockages
-country="$(curl --connect-timeout 10 -sSL 'https://geoip.kde.org/v1/ubiquity' | grep -Po '<CountryCode>\K([A-Z]{2})')"
+country="$(curl --connect-timeout 10 -sSL 'https://geoip.kde.org/v1/ubiquity' 2>/dev/null | grep -Po '<CountryCode>\K([A-Z]{2})')"
+
+# Fallback to ifconfig.co if KDE GeoIP is unavailable
+if [ -z "$country" ]; then
+    country="$(curl --connect-timeout 10 -sSL 'https://ifconfig.co/country-iso' 2>/dev/null | grep -Po '^[A-Z]{2}$')"
```
ventureoo, line 111:
```suggestion
    country="$(curl --connect-timeout 10 -sSL 'https://ifconfig.co/country-iso' 2>/dev/null')"
```
Grepping here seems unnecessary, it's already in ISO format.

ptr1337:
> Maybe use it by default since it's specialized service unlike `geoip.kde.org`?

Mhm, KDE.org would likely do less logging and information storage then ifconfig.co. But yeah, maybe we can change. I check their privacy policy

Issue #1147, opened 2026-02-20
Improved user experience of cachyos-rate-mirrors in China


lingdiansr:
Hi, I'm a Cahyos user from China, and I really appreciate the performance and ease-of-use improvements Cahyos has made. However, due to the Great Firewall of China, accessing international networks is often difficult. This causes a lot of trouble for me and other Chinese users, especially in terms of installation.

I noticed that `cachyos-rate-mirrors` uses `rate-mirrors` to optimize software sources. While I like this tool, its algorithm isn't well-suited for the network environment in mainland China.

I hope that `cachyos-rate-mirrors` can test mirror sources from mainland China only for `CN` users, thereby improving the ease of daily use and installation of the system.

In addition, [CDN77](https://github.com/CachyOS/CachyOS-PKGBUILDS/blob/6bf18aac13c195a3ac74b37cc09c2b687a09b4f8/cachyos-rate-mirrors/cachyos-rate-mirrors#L139)'s availability in mainland China is not high, so it is hoped that this mirror will not be inserted into `CN`.

**Solution**:
According to the help information for `rate-mirrors`, we can set `RATE_MIRRORS_COUNTRY_NEIGHBORS_PER_COUNTRY=0` to allow `CN` users to only test mirrors within their own country and not to explore neighboring mirrors, and set `RATE_MIRRORS_COUNTRY_TEST_MIRRORS_PER_COUNTRY=50` to test all mirrors within the country to ensure that multiple mirrors are available.
Regarding the CDN77 issue, i hope to add a condition in [here](https://github.com/CachyOS/CachyOS-PKGBUILDS/blob/6bf18aac13c195a3ac74b37cc09c2b687a09b4f8/cachyos-rate-mirrors/cachyos-rate-mirrors#L138) for determining the `CN` region to prevent the mirror from being inserted.

lingdiansr:
> I think we could try adding an equal workaround as we do with Russian country.

I created a PR #1148 , tested on my machine, able to pick only the Chinese mirror and not affect other regions.
Please check if there are any issues I may have overlooked.

PR #1148, opened 2026-02-20
Fix-cachyos-rate-mirrors-in-china

lingdiansr:
#1147


lingdiansr:
> Is china able to access the geoip.kde.org or https://ifconfig.co/country-iso

All can successfully obtain the country code.

PR #1152, opened 2026-02-21
github: Add issue template

ptr1337:
This adds issue templates for bugreports, package requests and package updates. Each on their own.

We **maybe** want to allow blank issues, but I think we could try it without and see how the reporters and users like it.

Its mostly templates and the user just selected them and types them in how they want.

They can be viewed here:
https://github.com/CachyOS/CachyOS-PKGBUILDS/blob/issue-template/.github/ISSUE_TEMPLATE/bug_report.yml

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -7,7 +7,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Please fill out this form as completely as possible. Incomplete reports are harder to act on.
+        Please fill out this form as completely as possible. Incomplete reports are harder to act on + If AI is used, please make sure that you're confident in its output. Bad reports will be closed and ignored.
 
   - type: input
     id: package
@@ -43,7 +43,7 @@ body:
     id: description
     attributes:
       label: Description
-      description: A clear and concise description of what the bug is.
+      description: A clear and concise description of what the bug is. Please avoid AI summaries as they might assume things that are not present.
     validations:
       required: true
 
--- a/.github/ISSUE_TEMPLATE/package_request.yml
+++ b/.github/ISSUE_TEMPLATE/package_request.yml
@@ -7,8 +7,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Before opening a request, please check that the package is not already present in this repository
-        or in the official Arch Linux / AUR repositories.
+        Before opening a request, please check that the package is not already present in this repository. Use https://packages.cachyos.org
 
   - type: input
     id: package


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,83 @@
...
+  - type: textarea
+    id: bugreport
+    attributes:
+      label: CachyOS bug report
+      description: |
+        Run `sudo cachyos-bugreport.sh` and paste the URL to the uploaded log here.
+        The script collects system information and uploads it automatically — only the link is needed.
+      placeholder: https://paste.cachyos.org/...
+    validations:
+      required: true
```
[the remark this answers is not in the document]
ptr1337: Yeah, youre right

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,85 @@
+name: Bug Report
+description: Report a broken or misbehaving package
+title: "[Bug]: "
+labels: ["bug"]
+assignees: [ptr1337, vnepogodin, 1Naim, ventueroo]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please fill out this form as completely as possible. Incomplete reports are harder to act on.
```
azdanov: `Please fill out this form as completely as possible. Incomplete reports are harder to act on.` + If AI is used, please make sure that you're confident in its output. Bad reports will be closed and ignored.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,85 @@
...
+        - Packaging error (wrong deps, missing files, etc.)
+        - Other
+    validations:
+      required: true
+
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: A clear and concise description of what the bug is.
```
azdanov: `A clear and concise description of what the bug is.` + Please avoid AI summaries as they might assume things that are not present.

.github/ISSUE_TEMPLATE/package_request.yml
```diff
@@ -0,0 +1,55 @@
+name: New Package Request
+description: Request a new package to be added to the repository
+title: "[Package Request]: "
+labels: ["package request"]
+assignees: [ptr1337, vnepogodin, 1Naim, ventueroo]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Before opening a request, please check that the package is not already present in this repository
```
azdanov: `the package is not already present`

Does this mean that if it's in AUR, then can't add? Phrasing of the whole sentence is confusing to me.

azdanov (approved):
:shipit:

.github/ISSUE_TEMPLATE/package_request.yml
ptr1337: More about to check if its already in packages.cachyos.org

PR #1169, opened 2026-02-26
mesa: one more try

ptr1337:
This follows the packaging we have in the "cachyos-testing" repository right now, which has been tested by users having issues with the yesterday's mesa.

It keeps the icd files for each architecutre.

The only diff, compared to the cachyos-testing package is that the rust depends has been updated.

### DIFF
--- a/mesa/mesa/PKGBUILD
+++ b/mesa/mesa/PKGBUILD
@@ -360,7 +360,7 @@ package_mesa() {
     _pick vkintel $libdir/libvulkan_intel{,_hasvk}.so
 
     _pick vkkosmic $libdir/libvulkan_kosmickrisp.so
-    _pick vkkosmic $icddir/kosmickrisp_mesa_icd.json
+    _pick vkkosmic $icddir/kosmickrisp_mesa_icd.*.json
 
     _pick vknvidia $icddir/nouveau_icd.*.json
     _pick vknvidia $libdir/libvulkan_nouveau.so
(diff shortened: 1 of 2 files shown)


mesa/mesa/PKGBUILD
```diff
@@ -123,7 +125,7 @@ declare -A _crates=(
   errno            0.3.12
   hashbrown        0.14.1
   indexmap         2.2.6
-  libc             0.2.168
+  libc             0.2.171
```
1Naim, line 128: Should stay at 0.2.168 https://gitlab.freedesktop.org/mesa/mesa/-/blob/staging/26.0/subprojects/packagefiles/libc-0.2-rs/meson.build?ref_type=heads#L7

mesa/mesa/PKGBUILD
```diff
@@ -349,6 +359,9 @@ package_mesa() {
     _pick vkintel $icddir/intel{,_hasvk}_icd.*.json
     _pick vkintel $libdir/libvulkan_intel{,_hasvk}.so
 
+    _pick vkkosmic $libdir/libvulkan_kosmickrisp.so
+    _pick vkkosmic $icddir/kosmickrisp_mesa_icd.json
```
1Naim: Don't we still have the arch dependent suffixes?

mesa/mesa/PKGBUILD
ptr1337: Mhm, thats questionable why arch set this. I look executing this weird script inside the pkgbuild

PR #1179, opened 2026-03-03
glibc: Disable sframe, add clearlinux.patch back

ptr1337:
We have removed in the recent incident issues the glibc version from us. This happened due -fno-semantic-interpostion was passed in docker-makepkg flags and then with a glib2 update this got an issue.

1. Add clearlinux.patch back with some minimal changes
2. Disable sframe to avoid issues with zig compilation, till archlinux enables it

Also, currently we do not provide anymore an "optimized" glibc for v3, v4 and znver4. I would add them back, since after removing the flag there is no issue anymore.


toolchain/glibc/PKGBUILD
```diff
@@ -15,7 +15,7 @@ pkgbase=glibc
 pkgname=(glibc lib32-glibc glibc-locales)
 pkgver=2.43+r5+g856c426a7534
 _commit=856c426a753450b8c6861a5b994a564f4fc16d4b
-pkgrel=2
+pkgrel=7
```
1Naim: Why is it this high? Repo is at -3 currently.

ptr1337: Ups, i did some local benchmarking

PR #1193, opened 2026-03-09
feat(cachyos-niri-noctalia): cleanup pkgbuild

azdanov:
Related: https://github.com/CachyOS/cachyos-niri-noctalia/pull/5

### DIFF
--- a/cachyos-niri-noctalia/.SRCINFO
+++ b/cachyos-niri-noctalia/.SRCINFO
@@ -22,7 +22,7 @@ pkgbase = cachyos-niri-noctalia
 	depends = xwayland-satellite
 	provides = cachyos-desktop-settings
 	conflicts = cachyos-desktop-settings
-	source = git+https://github.com/cachyos/cachyos-niri-noctalia#commit=911122088ceeb9bb9eed45965e725963303dd853
-	sha512sums = a99a0ab39f6f0640663eba685c3183ec23ed4883c4112e19a5479e55a73a4a3334ea6184051d07c54c663c6978e8114edf9fb213614e78a10f0cb3db0be3f140
+	source = git+https://github.com/cachyos/cachyos-niri-noctalia#tag=1.1.0
+	sha512sums = 4e19d7333fb99488984fc83169202acfba997d175a5800d0645465ac771823b9852dbed63b77bf93bb08161a8f40066911e2e107c3d03843bf494f2882cae142
 
 pkgname = cachyos-niri-noctalia
--- a/cachyos-niri-noctalia/PKGBUILD
+++ b/cachyos-niri-noctalia/PKGBUILD
@@ -10,7 +10,7 @@ license=(GPL-3.0-only)
 makedepends=('coreutils' 'git')
 install=$pkgname.install
 source=(git+$url#tag=${pkgver})
-sha512sums=('a99a0ab39f6f0640663eba685c3183ec23ed4883c4112e19a5479e55a73a4a3334ea6184051d07c54c663c6978e8114edf9fb213614e78a10f0cb3db0be3f140')
+sha512sums=('4e19d7333fb99488984fc83169202acfba997d175a5800d0645465ac771823b9852dbed63b77bf93bb08161a8f40066911e2e107c3d03843bf494f2882cae142')
 depends=(
   adw-gtk-theme
   cachyos-alacritty-config


cachyos-niri-noctalia/PKGBUILD
```diff
@@ -12,35 +12,24 @@ source=(git+$url#commit=911122088ceeb9bb9eed45965e725963303dd853)
...
+  capitaine-cursors
   niri
   noctalia-shell
-  xdg-desktop-portal-gnome
-  xwayland-satellite
-  pamixer
-  wl-clipboard
+  noto-fonts
+  noto-fonts-emoji
   polkit-kde-agent
```
azdanov: The `polkit-kde-agent` can be removed and replaced by https://noctalia.dev/plugins/polkit-agent/ which can be installed from the https://github.com/CachyOS/cachyos-niri-noctalia `skel` on initial user setup?

cachyos-niri-noctalia/PKGBUILD
```diff
@@ -12,35 +12,24 @@ source=(git+$url#commit=911122088ceeb9bb9eed45965e725963303dd853)
...
-  ttf-fantasque-nerd
-  ttf-font-awesome
-  otf-font-awesome
-  awesome-terminal-fonts
-  noto-fonts
-  noto-fonts-emoji
-  ttf-fira-sans
-  ttf-hack
-  kvantum
-  qt5ct
```
azdanov, line 25: I just found out that https://archlinux.org/packages/extra/x86_64/qt5ct/ has a hard dependency on https://archlinux.org/packages/extra/x86_64/qt5-wayland/
Otherwise it crashes. It probably worked before thanks to some transient link.

cachyos-niri-noctalia/PKGBUILD
azdanov: I moved `polkit-kde-agent` into optdepends to soft deprecate it, so it doesn't become an orphan and gets deleted accidentally.

cachyos-niri-noctalia/PKGBUILD
```diff
@@ -2,45 +2,35 @@
...
+  capitaine-cursors
   niri
   noctalia-shell
+  noto-fonts
+  noto-fonts-emoji
+  wl-clipboard
   xdg-desktop-portal-gnome
+  xorg-xwayland
```
azdanov: Not sure if `xorg-xwayland` is needed.

cachyos-niri-noctalia/.SRCINFO
```diff
@@ -1,40 +1,27 @@
...
-	depends = pamixer
-	depends = wl-clipboard
-	depends = polkit-kde-agent
-	depends = xorg-xwayland
-	depends = capitaine-cursors
-	depends = cachyos-wallpapers
-	depends = kvantum-theme-nordic-git
-	depends = brightnessctl
-	depends = cachyos-nord-gtk-theme-git
+	optdepends = polkit-kde-agent
```
1Naim: Since we have the polkit plugin from Noctalia, I don't think this is needed. If you want to keep this, add a description for it.

azdanov: Maybe can remove it completely. It's a big dependency for no special reason when noctalia can do same.

azdanov: I think I'll do a multi-step removal. At the moment just to opts and in a few months can be removed completely.

cachyos-niri-noctalia/PKGBUILD
```diff
@@ -2,45 +2,34 @@
...
 arch=('any')
 url="https://github.com/cachyos/$pkgname"
 license=(GPL-3.0-only)
 makedepends=('coreutils' 'git')
 source=(git+$url#commit=911122088ceeb9bb9eed45965e725963303dd853)
-sha512sums=('8696ec4c288a9db51d368c0d325dd5ce17ba2cfd2e87fecfb1c51b24f2faa864c82c80717a3469c174e4557483ba6a5141a04d080ac666f3db8b1031dce8b710')
+sha512sums=('a99a0ab39f6f0640663eba685c3183ec23ed4883c4112e19a5479e55a73a4a3334ea6184051d07c54c663c6978e8114edf9fb213614e78a10f0cb3db0be3f140')
 depends=(
   adw-gtk-theme
+  brightnessctl
```
1Naim: Can be removed, its a dependency of noctalia

cachyos-niri-noctalia/PKGBUILD
```diff
@@ -2,45 +2,34 @@
...
-  ttf-hack
-  kvantum
-  qt5ct
+  capitaine-cursors
   niri
   noctalia-shell
+  noto-fonts
+  noto-fonts-emoji
+  wl-clipboard
   xdg-desktop-portal-gnome
```
1Naim, line 24: We should also explicitly depend on xdg-desktop-portal-gtk

azdanov: https://packages.cachyos.org/package/extra/x86_64/xdg-desktop-portal-gnome
Is it needed since gnome depends on it?

1Naim: IMO yes. If someone ever chooses to branch out of the cachyos config and want to use a different portal other than gnome, removing gnome portal would also remove gtk portal. gtk portal should always be kept as fallback.

azdanov: Would they be able to remove it if it's an explicit dependency?

cachyos-niri-noctalia/cachyos-niri-noctalia.install
```diff
@@ -0,0 +1,8 @@
+pre_upgrade() {
```
1Naim:
```suggestion
post_upgrade() {
```

cachyos-niri-noctalia/cachyos-niri-noctalia.install
```diff
@@ -0,0 +1,8 @@
...
+  echo "==> NOTICE: polkit-kde-agent is no longer included by default."
+  echo "==> If necessary, explicitly reinstall 'polkit-kde-agent' or use the Noctalia polkit-agent plugin."
+}
+
+post_install() {
+  post_upgrade
+}
```
1Naim:
```suggestion
```
This doesn't work, and post_install isn't needed either.

cachyos-niri-noctalia/PKGBUILD
```diff
@@ -2,45 +2,33 @@
...
-pkgver=1.0.0
-pkgrel=4
+pkgver=1.1.0
+pkgrel=1
 arch=('any')
 url="https://github.com/cachyos/$pkgname"
 license=(GPL-3.0-only)
 makedepends=('coreutils' 'git')
+install=$pkgname.install
 source=(git+$url#commit=911122088ceeb9bb9eed45965e725963303dd853)
```
1Naim:
```suggestion
source=(git+$url#tag=${pkgver})
```

azdanov: Good catch. I was wondering what's up with that.

PR #1194, opened 2026-03-09
Add libjxl built with Sapphire Rapids and Zen4 codepaths

NexusXe:
For `libjxl`, `JPEGXL_ENABLE_AVX512_SPR` and `JPEGXL_ENABLE_AVX512_ZEN4` add optimized encoder codepaths for Sapphire Rapids and Zen 4, respectively. Like with the other AVX/AVX2/AVX3 codepaths already included in the build, these use runtime feature detection to determine which codepath to take.

Because the codepaths are dynamically selected, I think keeping the PKGBUILD simple and always including the optimized codepaths is a reasonable choice.

Below are some quick benchmarks of losslessly compressing [the first PNG I could find on Wikipedia](https://en.wikipedia.org/wiki/File:Diamond_%28side_view%29.png), running on an AMD AI 9 HX 370, with the command `time cjxl diamond.png diamond.jxl -q 100 -e $(EFFORT) -v -v`:
| Effort Level | Time Before | Time After |
|-|-|-|
|8|8.31|4.83|
|9|18.69|10.92|

In addition to the encoding speedups, the initial PNG decoding is over twice as fast as well (~90 MP/s to ~200 MP/s).

### DIFF
--- a/scx/scx-manager/.SRCINFO
+++ b/scx/scx-manager/.SRCINFO
@@ -1,6 +1,6 @@
 pkgbase = scx-manager
 	pkgdesc = Simple GUI for managing sched-ext schedulers via scx_loader
-	pkgver = 1.15.9
+	pkgver = 1.15.10
 	pkgrel = 1
 	url = https://github.com/cachyos/scx-manager
 	arch = aarch64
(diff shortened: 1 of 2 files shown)


libjxl/PKGBUILD
```diff
@@ -0,0 +1,125 @@
...
+    export CXXFLAGS+=' -DNDEBUG -ffat-lto-objects'
+    cmake -B build -S libjxl \
+        -G 'Unix Makefiles' \
+        -DCMAKE_BUILD_TYPE:STRING='None' \
+        -DCMAKE_INSTALL_PREFIX:PATH='/usr' \
+        -DCMAKE_POLICY_VERSION_MINIMUM:STRING='3.5.0' \
+        -DJPEGXL_BUNDLE_LIBPNG:BOOL='false' \
+        -DJPEGXL_ENABLE_AVX512:BOOL='true' \
+        -DJPEGXL_ENABLE_AVX512_SPR:BOOL='true' \
+        -DJPEGXL_ENABLE_AVX512_ZEN4:BOOL='true' \
```
1Naim, line 72: Is this the only diff between this and Arch's PKGBUILD? If yes, I don't see why the Arch PKGBUILD shouldn't enable this, since they are already enabling AVX512.

NexusXe: Yes, that is the only difference. I'm not sure why they don't. This change adds ~2 megabytes to the ~10 megabyte binary, maybe that's it? If you think it's likely to be accepted, I can open a PR on the Arch package instead.

ptr1337: Ill look into and ask the maintianer. I wonder if its an own libary or not. Will test.

PR #1207, opened 2026-03-12
cachyos-rate-mirrors: Enable timer to rank mirrors automatically

ptr1337:
Every two days, but maybe we can also increase to once a week?

Hook is also enabled to disable reflector in case enabled. I think for CachyOS we could generally remove it

### DIFF
--- a/cachyos-rate-mirrors/cachyos-rate-mirrors.install
+++ b/cachyos-rate-mirrors/cachyos-rate-mirrors.install
@@ -11,7 +11,3 @@ post_upgrade() {
 pre_remove() {
     systemctl disable --now cachyos-rate-mirrors.timer
 }
-
-post_remove() {
-    systemctl daemon-reload
-}
--- a/cachyos-rate-mirrors/.SRCINFO
+++ b/cachyos-rate-mirrors/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = cachyos-rate-mirrors
 	pkgdesc = CachyOS - Rate mirrors service
 	pkgver = 19
-	pkgrel = 2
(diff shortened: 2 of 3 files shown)


cachyos-rate-mirrors/cachyos-rate-mirrors.install
```diff
@@ -0,0 +1,17 @@
...
+    systemctl daemon-reload
+    systemctl enable --now cachyos-rate-mirrors.timer
+}
+
+pre_remove() {
+    systemctl disable --now cachyos-rate-mirrors.timer
+}
+
+post_remove() {
+    systemctl daemon-reload
```
ventureoo: This is not needed, it's already done by `30-systemd-daemon-reload-system.hook`

cachyos-rate-mirrors/cachyos-rate-mirrors.timer
ventureoo, line 1: I’m also wondering what `~` means here?

ventureoo: Nevermind, looks like it was GitHub glitch.

cachyos-rate-mirrors/cachyos-rate-mirrors.timer
```diff
@@ -3,7 +3,7 @@ Description=Automatically refresh arch / cachyos mirror lists
 
 [Timer]
 OnBootSec=15s
-OnUnitActiveSec=12h
+OnUnitActiveSec=2d
```
azdanov:
```suggestion
OnUnitActiveSec=5d
```

PR #1223, opened 2026-03-16
handheld: Add hhd-meta

1Naim:
Convenience meta package to switch to hhd from inputplumber

### DIFF
--- a/handheld/hhd-meta/PKGBUILD
+++ b/handheld/hhd-meta/PKGBUILD
@@ -1,0 +1,25 @@
+# Author : Manuel A. Fernandez Montecelo <mafm@igalia.com>
+# Maintainer: Eric Naim <dnaim@cachyos.org>
+
+pkgname='hhd-meta'
+pkgver=1.0.0
+pkgrel=1
+pkgdesc='Meta package for Handheld Daemon'
+arch=('any')
+depends=(
+  'cachyos-hooks'
+  'cachyos-settings'
+  'hhd'
+  'hhd-ui'
+)
+conflicts=('inputplumber' 'steamos-manager')
+install=hhd-meta.install
+source=('hhd-blacklist.conf'
+        'hhd-exports.sh')
+b2sums=('0042b4398212ee1360642fa898581d5cd54b87f964cda8dabaf311500353437fddb74d5240c44d556ab4d296cbcc9917bc6d8c0b7f301716b32eb1ca3a21b6b5'
+        '7727df8e5084c19a4eb9411bddef4a3318b29009592f02a2db7fbb610d3d04517d4e223e131cffa552eb17c7ce5c4a53b871b7b431ab96e0cfce9dca30720d66')
+
+package() {
+  install -Dm0644 hhd-blacklist.conf -t "${pkgdir}/usr/lib/modprobe.d"
+  install -Dm0755 hhd-exports.sh -t "${pkgdir}/etc/profile.d"
+}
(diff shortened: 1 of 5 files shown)


handheld/hhd-meta/hhd-exports.sh
```diff
@@ -0,0 +1,10 @@
+export HHD_PPD_MASK=1 HHD_HORI_STEAM=1 HHD_HIDE_ALL=1 HHD_BUGREPORT=1 HHD_FPASTE='paste-cachyos'
```
1Naim, line 1: These are actually for hhd.service and can be put in service override conf

handheld/hhd-meta/hhd-exports.sh
```diff
@@ -0,0 +1,10 @@
...
+060035ea0d0f00009601000001000000,Steam Controller (HHD),a:b0,b:b1,x:b2,y:b3,back:b6,guide:b8,start:b7,leftstick:b9,rightstick:b10,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,misc2:b11,
+060000000d0f00009601000002000000,Steam Controller (HHD),a:b0,b:b1,x:b2,y:b3,back:b6,guide:b8,start:b7,leftstick:b9,rightstick:b10,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,paddle1:b13,paddle2:b12,misc2:b11,
+060035ea0d0f00009601000003000000,Steam Controller (HHD),a:b0,b:b1,x:b2,y:b3,back:b6,guide:b8,start:b7,leftstick:b9,rightstick:b10,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,paddle1:b13,paddle2:b12,paddle3:b15,paddle4:b14,misc2:b11,
+030022faef170000ed61000010010000,Legion Go Dual DInput,crc:fa22,platform:Linux,a:b0,b:b1,x:b2,y:b3,leftx:a0,lefty:a1,lefttrigger:b4,righttrigger:b5,back:b7,start:b6,steam:2,
+$SDL_GAMECONTROLLERCONFIG
+END
+)
```
1Naim, line 10: This is supposed to be in gamescope-session but can't really modify that here so I bundled it with the service overrides. @ventureoo I'm putting this in /etc/profile.d right now, but do you know a better place to put it?

handheld/hhd-meta/PKGBUILD
```diff
@@ -0,0 +1,25 @@
...
+# Maintainer: Eric Naim <dnaim@cachyos.org>
+
+pkgname='hhd-meta'
+pkgver=1.0.0
+pkgrel=1
+pkgdesc='Meta package for Handheld Daemon'
+arch=('any')
+depends=(
+  'cachyos-hooks'
+  'cachyos-settings'
```
ptr1337, line 11: Is this needed?

1Naim: Depends on `paste-cachyos` and `update-initramfs` from those packages.

handheld/hhd-meta/hhd-exports.sh
ventureoo: Since it will be used only for handhelds, you can try to write our own systemd environment generator, see `man 7 systemd.environment-generator` and `/usr/lib/systemd/user-environment-generators/20-nvidia-rtd3-workaround` as example.

1Naim: Is it possible for 1 generator file to export multiple variables? The manpage doesn't tell me this and all examples that exist also only export 1 variable.

Also, theres system and user environment generators? Do I need to make 1 for each since hhd is a system service and gamescope-session is a user service?

ventureoo: > Is it possible for 1 generator file to export multiple variables? The manpage doesn't tell me this and all examples that exist also only export 1 variable.

Yes, it's possible, they're seperated by newline, e.g:

```fish
[N] ~ $ cat /etc/systemd/user-environment-generators/20-test-generator
#!/usr/bin/env sh
echo "TEST_VAR=1"
echo "TEST_SECOND_VAR=1"
[I] ~ $ systemctl --user show-environment | rg '^TEST'
TEST_VAR=1
TEST_SECOND_VAR=1
```
> Also, theres system and user environment generators? Do I need to make 1 for each since hhd is a system service and gamescope-session is a user service?

I think in this case you can write one generator and just install it in both directories.

1Naim: > I think in this case you can write one generator and just install it in both directories.

Hmmm, okay. I can just put the HHD env vars to its proper location then (service override) and put the SDL one in a generator.

PR #1266, opened 2026-03-26
snapper-support: Add helper to restore the installation to first snapshot

ptr1337:
This adds a desktop entry to restore the installation back to the installation point.

We do currently not snapshot home, so files will be preserved.
The Snapshot ID gets written to the filesystem, when the installer makes the final snapshot at the end of the installation.

The .config gets moved to the date of the reset to have a backup but not running into issues, when the user is resetting the installation.

I currently will use either kdialog or zenity for some information, when starting this script.

Just an idea right now, fully untested.

Depends on https://github.com/CachyOS/cachyos-calamares/pull/201


cachyos-snapper-support/cachyos-restore-installation
```diff
@@ -0,0 +1,56 @@
...
+
+SNAP_ID=$(cat "$SNAP_FILE")
+
+if [ -z "$SNAP_ID" ]; then
+    echo "Installation snapshot ID is empty." >&2
+    exit 1
+fi
+
+# Confirm before proceeding — this is a destructive, irreversible operation.
+CONFIRM_MSG="This will roll back to the original installation snapshot and reboot.\nAll changes since installation will be lost.\n\nContinue?"
```
ptr1337: Needs to add that documents and home stuff is preserved outside .config

cachyos-snapper-support/cachyos-restore-installation
```diff
@@ -0,0 +1,56 @@
...
+CONFIRM_TITLE="Restore Fresh Installation"
+
+confirmed=false
+if command -v kdialog > /dev/null 2>&1; then
+    kdialog --title "$CONFIRM_TITLE" --warningyesno "$CONFIRM_MSG" && confirmed=true
+elif command -v zenity > /dev/null 2>&1; then
+    zenity --question --title="$CONFIRM_TITLE" --text="$CONFIRM_MSG" --width=400 && confirmed=true
+else
+    echo "No GUI dialog tool found (kdialog or zenity required)." >&2
```
ventureoo: It also doesn't look right. If user doesn't have zenity or kdialog installed, they won't see anything at all when clicking on the entry. We should either make a hard dependency on one of them, or write our own graphical wrapper for it, maybe on Python/Lua using GTK/Qt.

PR #1309, opened 2026-04-14
Add a South Korea CachyOS mirror

kmw0410:
ROKFOSS is a FOSS mirroring project in South Korea and uses CDNs such as Cloudflare and Fastly

Mirror using **Cloudflare**:
- https://mirror.krfoss.org

Mirrors using **Fastly**:
- https://mirrors.krfoss.org
- https://mirror5.krfoss.org

Mirror **without a CDN**:
- https://mirror.distly.kr
- https://mirror.wane.kr


vnepogodin (changes requested):
sorry need rebase

PR #1329, opened 2026-04-22
wine-cachyos-opt: Fix permissions for xalia installation

topimiettinen:
Xalia zip as extracted produces files which are world writable. Fix the permissions afterwards.

### DIFF
--- a/wine-cachyos/PKGBUILD
+++ b/wine-cachyos/PKGBUILD
@@ -223,6 +223,8 @@ package() {
   cd "$srcdir"
   install -d -m755 "$pkgdir"/usr/share/xalia
   unzip xalia-${_xaliaver}-net48-mono.zip -d "$pkgdir"/usr/share/xalia
+  # Fix permissions
+  chmod -R go-w "$pkgdir"/usr/share/xalia
 }
 
 # vim:set ts=8 sts=2 sw=2 et:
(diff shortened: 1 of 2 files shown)


loathingKernel:
Nice find, thanks for the fix. Would you mind also doing it for `wine-cachyos` which also bundles xalia?

wine-cachyos/PKGBUILD
```diff
@@ -223,6 +223,8 @@ package() {
   cd "$srcdir"
   install -d -m755 "$pkgdir"/usr/share/xalia
   unzip xalia-${_xaliaver}-net48-mono.zip -d "$pkgdir"/usr/share/xalia
+  # Fix permissions
+  chmod -R go-w "$pkgdir"/opt/"${pkgname//-opt}"/share/xalia
```
loathingKernel: `wine-cachyos`, and thus xalia, are installed under a different path in this PKGBUILD.

loathingKernel (changes requested):

wine-cachyos/PKGBUILD
topimiettinen: oops

topimiettinen: Fixed, squashed and force pushed.

loathingKernel (approved):

loathingKernel:
Thank you, disregard the workflow, it fails for other reasons.

PR #1341, opened 2026-04-27
add a south korea cachyos mirror (updated)

kmw0410:
ROKFOSS is a FOSS mirroring project in South Korea and uses CDNs such as Cloudflare and Fastly.
`mirror5.krfoss.org` was already registered under a different address (hemino.net), so I updated the information.

Mirror using **Cloudflare**:
- https://mirror.krfoss.org/
- https://mirror4.krfoss.org/

Mirrors using **Fastly**:
- https://mirrors.krfoss.org/

**Update: some mirrors that do not use CDN have been removed or changed due to potential issues**


vnepogodin (approved):
Many thanks

PR #1361, opened 2026-05-03
python: enable tail-call interpreter with GCC 16

ptr1337:


ptr1337:
https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/1216

Here is more explaination. GCC 16 now does support it.

PR #1366, opened 2026-05-06
nvidia: add patches to fix rainbow issue on certain vr headsets

nootsponge:
Following the conversation here: https://discord.com/channels/862292009423470592/871363526541852692/1482971455842025523 (post by @1Naim I think?)

I've been using these patches forever and I'm frankly tired of having to reapply them with every update, and as far as I can tell it was an intention to get these patched for good, so I'm trying to get the ball rolling.

If I need to condense the patches into one file, or do anything else really, let me know. I'm unfamiliar with contributing here so advice is appreciated.


1Naim (approved):
Please also bump the pkgrel and regenerate .SRCINFO.

1Naim (approved):
LGTM. Since the prebuilt modules already have this I think it's fine to merge this right away.

PR #1377, opened 2026-05-11
rate-mirrors: remove vendored openssl and re-enable LTO

Integral-Tech:
Link against the system-provided openssl instead of the bundled copy.

After removing vendored openssl, link-time optimization (LTO) can be enabled again.


ptr1337 (approved):

ptr1337:
Thanks!

PR #1393, opened 2026-05-17
nodejs: remove vendored libffi

Integral-Tech:
Link against the system-provided libffi instead of the bundled copy.


ptr1337 (approved):
Thanks!

PR #1433, opened 2026-05-29
nodejs: enable LTO

Integral-Tech:
LTO (Link Time Optimization) has been enabled for `nodejs` in the Arch Linux official repository,


ptr1337:
Oh, thanks. Thought we have been using ThinLTO, but did not :D

PR #1437, opened 2026-05-30
proton-cachyos: version 11.0-20260521

loathingKernel:
- **proton-cachyos: version 11.0-20260521**
- **proton-cachyos-slr: version 11.0-20260521**


ptr1337 (approved):
Yaay, got more then time :P

PR #1440, opened 2026-06-02
Add mirror.hjk.gg to mirrorlist

hjk321:
Syncs from tier 0 via syncthing. I can be contacted at any time via GitHub message or on discord as hjk321.


vnepogodin (approved):
Many thanks

Issue #1463, opened 2026-06-09
[Bug]: cachyos-rate-mirrors service not being trigged by timer


Mattyan89:
### Package name

cachyos-rate-mirrors

### Package version

23-2

### Issue type

Runtime / functionality issue

### Description

I think the service for cachyos-rate-mirrors is never trigged, at least on a new install.
There is no programmed trigger in the timer:
<img width="987" height="187" alt="Image" src="https://github.com/user-attachments/assets/aa2eb89d-3713-4791-ad3e-13516805604e" />
Unless you manually start the service first:
<img width="996" height="176" alt="Image" src="https://github.com/user-attachments/assets/1cfa2f37-9292-44b4-876e-8d29ef21f2af" />
then the trigger is correctly programmed for 10 days later.
OnUnitActiveSec requires that the service has run at least once before, otherwise it can't be programmed.

### Steps to reproduce

1. Do a fresh install
2. systemctl status cachyos-rate-mirrors.timer (notice how the trigger isn't set)
3. systemctl start cachyos-rate-mirrors.service
4. Now the trigger is set and the service will correctly execute in due time

### Relevant logs or error output

```shell

```

PR #1470, opened 2026-06-12
[cachyos-rate-mirrors] Fix timer never scheduling on fresh install

Mattyan89:
I would like to suggest a solution to #1463.
The current OnUnitActiveSec is ineffective if the service hasn't run at least once before.
For what I see, on a fresh install rate-mirrors.service does not run so it is never scheduled.
OnCalendar would instead schedule the service on the 10th 20th 30th every month, while RandomizedOffsetSec sets a fixed delay between 0 and 10 days so clients are distributed each day. You'll have some machines scheduled to update eg. on the 14th,24th,4th (or 3rd), others on the 15th,25th,5th (or 4th) and so on.
RandomizedOffsetSec is better than RandomizedDelaySec for this use case as systemd docs says: "this setting is most useful to prevent resource congestion on a remote service, from a fleet of similarly-configured clients"
This should fix the issue while keeping the refresh interval roughly the same.


Mattyan89:
Fixed, I hadn't thought about the fact that oncalendar obviously requires real dates.
If I read the documentation correctly this time, randomized should then be able to calculate a valid date.

ptr1337 (approved):

PR #1475, opened 2026-06-13
Use heroic launcher binary

Shagon94:
The `heroic-games-launcher` is built using a lot of dependencies, including npm. There's already a prebuilt package that we can use.


witchlliee:
duplicate #1464

PR #1489, opened 2026-06-14
melonds: Add at v1.1

### LINKED ISSUE #1485: [Package Request]: MelonDS, a DS emulator, sorta
### Package name

melonds

### Upstream / source URL

https://melonds.kuribo64.net/

### Current availability

AUR only

### Description

I would really appreciate if this DS emulator got into CachyOS repos, the future of AUR security is uncertain, so making sure packages related to gaming (aka:. things a regular gamer probably would install) is on the main repos is crucially important.

### Additional context

https://aur.archlinux.org/packages/melonds <-- This is the AUR page, this is a compilation PKGBUILD to get melonds installed on your system.

witchlliee:
Resolves #1485

### DIFF
--- a/melonds/.SRCINFO
+++ b/melonds/.SRCINFO
@@ -1,0 +1,27 @@
+pkgbase = melonds
+	pkgdesc = DS emulator, sorta
+	pkgver = 1.1
+	pkgrel = 1
+	url = http://melonds.kuribo64.net/
+	arch = x86_64
+	license = GPL-3.0-or-later
+	makedepends = wayland
+	makedepends = git
+	makedepends = cmake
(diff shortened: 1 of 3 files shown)


ptr1337:
Im kinda afraid, that this could result into a lawsuit from nintendo?

witchlliee:
Yeah I don't understand much about that. The emulator seems fine to distribute but you never know when talking about nintendo.

vnepogodin:
wouldn't want to distribute any nintendo emulators as we aren't allowed. keeping pkgbuild in that repo neither make sense if we don't distribute it in production

PR #1491, opened 2026-06-15
aura: Update to 1.2.0

witchlliee:
Does it even the patch anymore? I tested it here and it seems to work fine without it at the moment.


vnepogodin (approved):
LGTM

PR #1502, opened 2026-06-16
cachyos-handheld: Move bootstrap service into systemd one shot service

ptr1337:
The .install has now a restricted network access. This results into that bootstrasp is not working. Move to systemd service


handheld/cachyos-handheld/cachyos-handheld-bootstrap.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+# Download and install the SteamOS bootstrap to speed up first boot setup.
+# Runs as a first-boot oneshot service because pacman no longer permits
+# network access from package scriptlets (.install).
+set -euo pipefail
+
+URL="https://steamdeck-packages.steamos.cloud/archlinux-mirror/jupiter-main/os/x86_64/steam-jupiter-stable-1.0.0.85-2-x86_64.pkg.tar.zst"
```
ventureoo, line 7: Is this URL persistent through? What if Valve will change package version?

handheld/cachyos-handheld/cachyos-handheld-bootstrap.sh
```diff
@@ -0,0 +1,28 @@
...
+BOOTSTRAP_SHA512=a013473d28a9d10a865b543ee1c42fa514c4af383b63f7dbf8e38d18e358ebdbc247caff1e39cc372a71b05fdd506f2356e1af0c5404a95f6615191d994e0fbf
+DEST=/usr/lib/steam/bootstraplinux_ubuntu12_32.tar.xz
+
+workdir=$(mktemp -d)
```
ventureoo, line 11: You can stream content of the archive with pipe instead of using temp file.

ptr1337: Ive just copied the content of the script as it was before. We can do further adjustments when the network stuff is fixed.

ventureoo:
Can also someone tell me (maybe @1Naim) why we can't just repackage or add to our repo steam-jupiter-stable package and install it just as is?

ptr1337:
> Can also someone tell me (maybe @1Naim) why we can't just repackage or add to our repo steam-jupiter-stable package and install it just as is?

That **could** work too.

ptr1337:
We need to move it into Calamares Shellprocess. The current way just makes the whole bootstrap uselesss, since it considers to improve startup time.

1Naim:
> Can also someone tell me (maybe @1Naim) why we can't just repackage or add to our repo steam-jupiter-stable package and install it just as is?

We can, and I've suggested this as well.

PR #1506, opened 2026-06-17
steamos-jupiter-stable: Introduce package for oob bootstrap

ptr1337:


handheld/steam-jupiter-stable/70-steam-jupiter-leds.rules
1Naim, line 1: cachyos-handheld provides a similar rule for this, so let's just leave this out. It doesn't work anyway because user:group is hardcoded to `deck`

handheld/steam-jupiter-stable/70-steam-jupiter-input.rules
1Naim, line 1: I guess we can include this but IIRC this was discussed for cachyos-handheld package too and we decided not to include it.

1Naim (approved):
Looks good overall for the short term solution. Please also remove the bootstrap setup in `cachyos-handheld.install`

handheld/steam-jupiter-stable/PKGBUILD
```diff
@@ -0,0 +1,137 @@
...
+  lib32-libxcrypt-compat
+  lib32-libgl
+  lib32-gcc-libs
+  lib32-libx11
+  lib32-libxss
+  lib32-alsa-plugins
+  lib32-libgpg-error
+  lib32-nss
+  lib32-vulkan-driver
+  lib32-vulkan-icd-loader
```
ventureoo, line 65: We need to remove vulkan-driver dependency here, or split this package to avoid pulling mesa-git at netinstall step in calamares, since it creates conflict with chwd profile later. Spotted during testing https://github.com/CachyOS/cachyos-calamares/pull/168: https://paste.cachyos.org/f0cfa09.log

PR #1513, opened 2026-06-19
proton-cachyos-{native,slr}: version 11.0-20260602-3

loathingKernel:
* Removed `proton-cachyos`
* Added `proton-cachyos-native` providing `proton-cachyos` and replacing `proton-cachyos`
* Kept `proton-cachyos-slr` as-is providing `proton-cachyos` and replacing `proton-cachyos`

This will need for the `proton-cachyos` package to be removed from the repos to work.

Removed `lib32-lzo` dependency coming from AUR which can be dropped if it's not needed by anything else. Added a few 32bit dependencies from the repos.

Todo: wiki package names

Drafted until reviewed for the packaging changes


loathingKernel:
On second thought, anyone that has installed `proton-cachyos` until now, already knows about it. So let's transition them to the new `proton-cachyos-native`package. Both `proton-cachyos-native` and `proton-cachyos-slr` have `replaces=proton-cachyos` to prepare for an eventual removal of the native one if it comes to that.

loathingKernel:
@ptr1337 bump

PR #1519, opened 2026-06-20
Set paru's RepoOnly option

namesarebloat:
Arch, which is significantly less often recommend to and used by linux beginners, won't even allow AUR helpers in its repos, let alone install one by default. CachyOS should put up at least this tiny barrier between the AUR and people who absolutely shouldn't get their packages from it. Not sure if that's the correct way to do that though.

### DIFF
--- a/paru/PKGBUILD
+++ b/paru/PKGBUILD
@@ -59,4 +59,5 @@ package() {
 
   install -d "$pkgdir/usr/share/"
   cp -r locale "$pkgdir/usr/share/"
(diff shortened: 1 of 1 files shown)


1Naim:
This defeats the purpose of shipping `paru` installed by default. If we want to have this 'tiny barrier', then we should not install an AUR helper at all.

simiscoool-afk:
paru is not installed by default and has been replaced by shelly which brings with it a extra warning to enable the AUR and some basic pkgbuild scanning.

1Naim:
Then there's even less purpose for this PR. Closing.

PR #1527, opened 2026-06-23
[cachyos-rate-mirrors] RandomizedDelaySec + more elegant synthax 

Mattyan89:
As it is now, if the system is off when the timer would have fired, the service starts immediately after boot. 
Re-added RandomizedDelaySec (since this option is carried to the next boot) to further delay it (1h maximum) and avoid any problems.
Also, cleaner synthax for OnCalendar (funcionally the same).


Mattyan89:
I messed up, I'll make a new PR

PR #1537, opened 2026-06-26
cachyos-settings: migrate X11 keyboard layout via localectl

ptr1337:
Move existing /etc/X11/xorg.conf.d/00-keyboard.conf keymap settings into systemd-localed when the localectl X11 layout is still unset. Run the migration on install and upgrade, and mark it as completed afterward.

### DIFF
--- a/cachyos-settings/cachyos-settings.install
+++ b/cachyos-settings/cachyos-settings.install
@@ -14,7 +14,7 @@ migrate_keyboard_layout() {
 
     layout=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbLayout"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
     model=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbModel"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
-    variant=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbVariant"/ { print $4; exit }' "$x11_conf")
+    variant=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbVariant"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
     options=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbOptions"/ { print $4; exit }' "$x11_conf")
 
     [[ -n "$layout" ]] || return 0
--- a/cachyos-settings/cachyos-settings.install
+++ b/cachyos-settings/cachyos-settings.install
@@ -15,7 +15,7 @@ migrate_keyboard_layout() {
     layout=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbLayout"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
     model=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbModel"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
     variant=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbVariant"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
-    options=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbOptions"/ { print $4; exit }' "$x11_conf")
+    options=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbOptions"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
 
     [[ -n "$layout" ]] || return 0
     [[ -n "$model" ]] || model="pc105"
--- a/cachyos-settings/cachyos-settings.install
+++ b/cachyos-settings/cachyos-settings.install
@@ -20,7 +20,7 @@ migrate_keyboard_layout() {
     [[ -n "$layout" ]] || return 0
     [[ -n "$model" ]] || model="pc105"
 
-        localectl set-x11-keymap "$layout" "$model" "$variant" "$options" || return 0
+    localectl set-x11-keymap "$layout" "$model" "$variant" "$options" || return 0
     elif [[ -n "$variant" ]]; then
         localectl set-x11-keymap "$layout" "$model" "$variant" || return 0
     else
--- a/cachyos-settings/cachyos-settings.install
+++ b/cachyos-settings/cachyos-settings.install
@@ -22,7 +22,6 @@ migrate_keyboard_layout() {
     localectl set-x11-keymap "$layout" "$model" "$variant" "$options" || return 0
         localectl set-x11-keymap "$layout" "$model" "$variant" || return 0
     else
-    fi
 
     install -Dm644 /dev/null "$marker" 2>/dev/null || return 0
 }
--- a/cachyos-settings/cachyos-settings.install
+++ b/cachyos-settings/cachyos-settings.install
@@ -42,5 +42,7 @@ post_upgrade() {
             systemctl restart "$service"
         fi
     done
-    migrate_keyboard_layout
+    if (( $(vercmp "$2" 1:1.3.5-2) < 0 )) && (( $(vercmp "$2" 1:1.3.5-1) >= 0 )); then
+        migrate_keyboard_layout
+    fi
 }
--- a/cachyos-settings/cachyos-settings.install
+++ b/cachyos-settings/cachyos-settings.install
@@ -20,7 +20,6 @@ migrate_keyboard_layout() {
     [[ -n "$layout" ]] || return 0
 
     localectl set-x11-keymap "$layout" "$model" "$variant" "$options" || return 0
-        localectl set-x11-keymap "$layout" "$model" "$variant" || return 0
 
     install -Dm644 /dev/null "$marker" 2>/dev/null || return 0
 }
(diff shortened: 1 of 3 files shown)


ptr1337:
@vnepogodin Please review as soon as possible to avoid further interuption.

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+    local status layout model variant options
+
+    [[ -e "$marker" ]] && return 0
+    [[ -r "$x11_conf" ]] || return 0
+    status=$(localectl status 2>/dev/null) || return 0
+    if ! grep -q '^[[:space:]]*X11 Layout:[[:space:]]*(unset)' <<< "$status"; then
+        return 0
+    fi
+
+    layout=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbLayout"/ { print $4; exit }' "$x11_conf")
```
vnepogodin:
```suggestion
    layout=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbLayout"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+    model=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbModel"/ { print $4; exit }' "$x11_conf")
```
vnepogodin:
```suggestion
    model=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbModel"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+    variant=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbVariant"/ { print $4; exit }' "$x11_conf")
```
vnepogodin:
```suggestion
    variant=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbVariant"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+    options=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbOptions"/ { print $4; exit }' "$x11_conf")
```
vnepogodin:
```suggestion
    options=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbOptions"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+
+    [[ -n "$layout" ]] || return 0
+    [[ -n "$model" ]] || model="pc105"
```
vnepogodin:
```suggestion
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+
+    if [[ -n "$options" ]]; then
```
vnepogodin:
```suggestion
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+        localectl set-x11-keymap "$layout" "$model" "$variant" "$options" || return 0
```
vnepogodin:
```suggestion
    localectl set-x11-keymap "$layout" "$model" "$variant" "$options" || return 0
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+    elif [[ -n "$variant" ]]; then
```
vnepogodin:
```suggestion
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+        localectl set-x11-keymap "$layout" "$model" "$variant" || return 0
```
vnepogodin:
```suggestion
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+    else
```
vnepogodin:
```suggestion

```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+        localectl set-x11-keymap "$layout" "$model" || return 0
```
vnepogodin:
```suggestion
```

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,36 @@
...
+    fi
```
vnepogodin:
```suggestion

```

vnepogodin (commented):
I really don't like like using these ugly awk parsers

ptr1337:
> I really don't like like using these ugly awk parsers

How you want to have it done?

vnepogodin:
> > I really don't like like using these ugly awk parsers
> 
> How you want to have it done?

application/lib with such functionality instead of doing script parsing via awk. which is unrelated for our hot patch workaround

ventureoo:
Why do we need this in first place?

ptr1337:
> Why do we need this in first place?

xkeyboard-config did broke all current set layouts. Users are booting their machines and do not have any layout set and defaulting to US Layout.

This resulting into that many users can not login anymore, its also mentioned in commit msg

ptr1337:
We got alone since today morning likely 30-50 reports about this issue.

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,29 @@
...
+    local status layout model variant options
+
+    [[ -e "$marker" ]] && return 0
+    [[ -r "$x11_conf" ]] || return 0
+    status=$(localectl status 2>/dev/null) || return 0
+    if ! grep -q '^[[:space:]]*X11 Layout:[[:space:]]*(unset)' <<< "$status"; then
+        return 0
+    fi
+
+    layout=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbLayout"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
```
ventureoo, line 13: Can't we use `setxkbmap -query` instead of awk on Xorg config?

vnepogodin:
well we can't use localectl for new installs as it needs dbus

cachyos-settings/cachyos-settings.install
ptr1337: No, this does not work, when there is no x11 enviroment available

Also, we want to migrate the legacy config

ptr1337: Also tested right now on my machine and this does not work. It returns US Layout:
```
~/p/c/CachyOS-Live-ISO master• 28m 37.8s ❱ setxkbmap -query
WARNING: Running setxkbmap against an Xwayland server
rules:      evdev
model:      pc105
layout:     us
```

ventureoo: > No, this does not work, when there is no x11 enviroment available
> 
> Also, we want to migrate the legacy config

The issue is that you are quering only one fixed path configuration, which doesn’t take into account other user settings.

ptr1337:
> well we can't use localectl for new installs as it needs dbus

New installs are not affected, since it only runs on post_upgrade. We can pin it a a specific cachyos-settings version

cachyos-settings/cachyos-settings.install
ptr1337: Well, it depends on the installation how we had it previously. I dont think users did overwrite this config at all

vnepogodin:
> > well we can't use localectl for new installs as it needs dbus
> 
> New installs are not affected, since it only runs on post_upgrade. We can pin it a a specific cachyos-settings version

ah you updated, for ventureoo context anyways

cachyos-settings/cachyos-settings.install
```diff
@@ -1,5 +1,29 @@
...
+    model=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbModel"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
+    variant=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbVariant"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
+    options=$(awk -F'"' '/^[[:space:]]*Option[[:space:]]+"XkbOptions"/ { gsub(/^[[:space:]]+|[[:space:]]+$/, "", $4); print $4; exit }' "$x11_conf")
+
+    [[ -n "$layout" ]] || return 0
+
+    localectl set-x11-keymap "$layout" "$model" "$variant" "$options" || return 0
+
+    install -Dm644 /dev/null "$marker" 2>/dev/null || return 0
```
ventureoo: I don’t think the marker is needed here, as we still only do it in case of unset

ptr1337: Yep, ill remove it. Was intended for new installations to add a marker, when calamares is fixed and rolled out, but since we can fix existing calamares anyways this can be removed.

cachyos-settings/cachyos-settings.install
ventureoo: Honestly, not a big fun of such workarounds, I would like rather revert broken upstream changes. But since issue is urgent, then fine.

ptr1337: This is not upstream broken, this is an intended change without migration path

55f536539197e25373c94dc7b8a2e80eb3548551
/usr/share/X11/xkb --> /usr/share/xkeyboard-config-2

ventureoo: So, we can’t just re-package it by adding a symlink, right?

ptr1337: I dont know, but following upstream would be better there. Tomorrow we plan to make the release and new installation are fixed too with new calamares.

We just need to migrate existing users properly, which this makes it.

PR #1546, opened 2026-06-28
cachyos-hypr-noctalia: removing some deps

Jinra:
* removed sddm to not force users to use it. instead offering it as a default dm via calamares netinstall
* removed some other deps as they're duplicated in the base install and not needed for this package


Jinra:
@ptr1337 can we please merge this before the next official iso build

PR #1560, opened 2026-07-05
cachyos-hypr-noctalia: Noctalia v5 support

Jinra:
This PR updates the package to Noctalia v5


Jinra:
@ptr1337 :pray:

ptr1337:
How do we ensure the migration works? The dotfiles are in /etc/skel so users would need to migrate manually i guess

azdanov:
From what I understand v4 and v5 are independant. So can include both + post_install/update echo telling users to manually upgrade. And after few weeks remove v4?

witchlliee:
> From what I understand v4 and v5 are independant. So can include both + post_install/update echo telling users to manually upgrade. And after few weeks remove v4?

Oh I guess this will cause problems on my dots as well. Since I'm just replacing. I guess I can just include post_upgrade echo recommending manually upgrade? Still kinda confused how pacman will handle this. `cachyos-mango-dms` remains there but it points to a git where I already changed to noctalia.

Jinra:
> How do we ensure the migration works? The dotfiles are in /etc/skel so users would need to migrate manually i guess

@ptr1337 I added migration steps on the wiki (going to add to the Discord #guide as well). I also added an install message in the PKGBUILD for when users upgrade directing them to follow the directions on the wiki. However, I'm not sure if it's 100% correct, if you could verify for me that would be great.

wiki guide: https://wiki.cachyos.org/configuration/desktop_environments/hyprland/#migrating-from-v4-dots-to-v5

Jinra:
@ptr1337 _ping_ :pray:

ptr1337 (approved):

PR #1578, opened 2026-07-08
Correct format for CachyOS mirrorlist

ParadaCarleton:
### Description
This PR corrects the mirror country comment headers in `cachyos-mirrorlist`. 

Currently, `rate-mirrors` (which is called by `cachyos-rate-mirrors`) parses countries by looking for standard double-hash country headers directly preceding the mirror entries (e.g., `## United States` or `## Germany`). 

Because the current list uses custom credit strings (e.g., `## USA Mirror much thanks to corpdecker!`) and `## tier=2 code=US` on the `##` lines, `rate-mirrors` is unable to associate any CachyOS mirrors with their respective countries. This forces the tool to query all 28 mirrors as "unlabeled" regardless of user country configurations or neighbor limits.

### Changes Made
- Rewrote the `##` lines to contain only the official country name (which `rate-mirrors` maps to ISO codes).
- Moved the tier levels, codes, and credit notes to single-hash comments (`#`) directly below the country name to preserve existing metadata and contributor credits without confusing parser regex engines.

Let me know if I made any mistakes; I'm not sure how I'd add tests for this.

### DIFF
--- a/cachyos-mirrorlist/cachyos-mirrorlist
+++ b/cachyos-mirrorlist/cachyos-mirrorlist
@@ -3,115 +3,154 @@
 ####        CachyOS Repository Mirrorlist         ####
 ####                                              ####
 ######################################################
-## CDN77 (World Wide Datacenters)
-## tier=1 code=GLOBAL
+## Worldwide
+# CDN77 (World Wide Datacenters) | tier=1 code=GLOBAL
 Server = https://cdn77.cachyos.org/repo/$arch/$repo
-## Rerouted to mirror
-## tier=1 code=GLOBAL
+
+## Worldwide
+# Rerouted to mirror | tier=1 code=GLOBAL
 # Server = https://cdn.cachyos.org/repo/$arch/$repo
-## Germany, disabled due Netcup issues:
-## : OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0"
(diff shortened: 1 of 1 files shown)


ParadaCarleton:
Mentioning @ptr1337 @vnepogodin @sirlucjan —PR should be quick to review and installing packages from untrusted (by the user) mirrors is a security hole.

vnepogodin:
No, its correct and follows new API spec

ParadaCarleton:
> No, its correct and follows new API spec

I've tested it and, even after setting the configuration in `/etc/systemd/system/cachyos-rate-mirrors.service.d/override.conf` to use US only, pacman continues to use Russian mirrors:
```
[Service]
Environment="RATE_MIRRORS_ENTRY_COUNTRY=US"
Environment="RATE_MIRRORS_COUNTRY_NEIGHBORS_PER_COUNTRY=0"
```
So something is _definitely_ wrong with either the cachyos-rate-mirrors script or with the format.

vnepogodin:
New API isn't used in rate-mirrors yet

ParadaCarleton:
> New API isn't used in rate-mirrors yet

In that case, the mirrorlist configuration shouldn't be updated until the API has been corrected for rate-mirrors.

PR #1585, opened 2026-07-13
asusctl: update repository location

v1993:
The asusctl project has been moved to https://github.com/OpenGamingCollective/asusctl. No new versions were published in the new location as of now, so this just updates the URL.

### DIFF
--- a/asusctl/.SRCINFO
+++ b/asusctl/.SRCINFO
@@ -16,7 +16,7 @@ pkgbase = asusctl
 	makedepends = rust
 	makedepends = seatd
 	makedepends = systemd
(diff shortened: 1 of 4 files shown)


ptr1337:
I think we could move the whole PKGBUILD to cachyos-aur-derived.

v1993:
What's the exact purpose of that repository? It would be nice to have documentation on what goes where, but it currently doesn't even have a readme.

AFAIK, the asusctl was added here because the AUR package took a long time to update.

vnepogodin (approved):

v1993:
As a side note - asusctl actually got two releases since when I opened this PR.

vnepogodin:
known. updated already

PR #1677, opened 2026-08-05
fix(alsa-ucm-conf): apply upstream patch for Turkish locale

### LINKED ISSUE #533: [Bug]: Upstream alsa-ucm-conf regression affecting Turkish locale on Intel Meteor Lake SOF/SoundWire
### Kernel

7.1.5-1-cachyos

### Desktop Environment / Window Manager

KDE Plasma

### CPU

Intel Core Ultra 7 155H

### GPU / Driver

Intel Arc Graphics / xe

## Summary

This is **not a CachyOS-specific bug**, but it affects CachyOS users.

Initially the installed system exposed only **Dummy Output**, while the CachyOS Live ISO worked perfectly on the same hardware.

After a full investigation, the root cause was identified as an upstream `alsa-ucm-conf` bug.

Upstream issue:

https://github.com/alsa-project/alsa-ucm-conf/issues/824

---

## Affected hardware

- Acer Swift SFG14-72
- Intel Meteor Lake-P HD Audio (8086:7e28)
- SOF + SoundWire

---

## Root cause

The issue is caused by a locale-dependent POSIX regular expression in `sof-soundwire.conf`.

Under `tr_TR.UTF-8`, ALSA fails to import the UCM profile. As a result, PipeWire/WirePlumber only expose the `off` profile and the desktop shows **Dummy Output**.

The problem is **not** related to PipeWire, WirePlumber, the kernel, firmware or CachyOS packaging itself.

---

## Suggestion

Until the upstream fix is merged and released, CachyOS could consider carrying the small upstream-compatible patch discussed in the ALSA issue for affected users.

I am happy to test candidate patches or provide additional logs if needed.

Thank you.

### Steps to Reproduce

1. Install CachyOS on an Intel Meteor Lake SOF/SoundWire laptop.
2. Use the Turkish locale (`tr_TR.UTF-8`).
3. Log into KDE Plasma.
4. Open the audio settings or run `pactl list cards`.

The system exposes only `Dummy Output` because the HiFi profile is missing.

The same hardware works correctly when booting the CachyOS Live ISO.

### Expected Behavior

The installed system should expose the normal HiFi profile, speakers, headphones and internal microphone exactly like the CachyOS Live ISO.

### Actual Behavior

The installed system exposes only `Dummy Output`.

The HiFi profile is missing because ALSA fails to import the SOF SoundWire UCM configuration under `tr_TR.UTF-8`.

The root cause has been identified as an upstream `alsa-ucm-conf` issue:

https://github.com/alsa-project/alsa-ucm-conf/issues/824

### Logs / Error Messages

No CachyOS-specific logs.  The issue has been isolated to an upstream alsa-ucm-conf bug.  See upstream report: https://github.com/alsa-project/alsa-ucm-conf/issues/824

### Additional Context

The investigation confirmed that:

- The same hardware works correctly with the CachyOS Live ISO.
- The installed system fails only under `tr_TR.UTF-8`.
- The issue has been fully root-caused and reported upstream.
- A temporary downstream patch could be carried until the upstream fix is released.

Upstream issue:

https://github.com/alsa-project/alsa-ucm-conf/issues/824

SisyphusAeolides:
Backports upstream regex fix from alsa-project/alsa-ucm-conf#824 to resolve Dummy Output issues on Meteor Lake hardware under tr_TR.UTF-8.

Resolves CachyOS/distribution#533

### DIFF
--- a/core/alsa-ucm-conf/PKGBUILD
+++ b/core/alsa-ucm-conf/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer: David Runge <dvzrv@archlinux.org>
+# Contributor: Jan Alexander Steffens (heftig) <jan.steffens@gmail.com>
+
+pkgname=alsa-ucm-conf
+pkgver=1.2.16.1
+pkgrel=2
+pkgdesc="ALSA Use Case Manager configuration (and topologies)"
+arch=(any)
+url="https://alsa-project.org/"
+license=(BSD-3-Clause)
+source=(
+  $url/files/pub/lib/$pkgname-$pkgver.tar.bz2{,.sig}
+  fix-turkish-locale.patch
+)
+sha512sums=('df25a29afc3b87347fa89f3fe60ebda68068acdaf68e081b58455ebee319c21e82af4076ba1f10781ba620a9e8efa6e65db27c5989d63a429d44a715dbd1d460'
+            'SKIP'
+            'c52f017b419b8565fc331738938b72aeb2374cac867b06065fcc609fa5eb93c16b3c453d4d0c039617af387afa14f9feed028e210cd77320409afdbb41f3a534')
+b2sums=('753e1775cd99c0f6289507d04760c7a87ead4a0d029a817aa3e6083a30eae123859a91432511f1e5f4daa25ddee960e8f3be130586a2761ee95bf4af40a76a13'
+        'SKIP'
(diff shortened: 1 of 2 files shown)


ptr1337:
Wouldnt this be not better that upstream handles it?

SisyphusAeolides:
> Wouldnt this be not better that upstream handles it?

It is already fixed upstream. This PR backports the targeted upstream regex fix so CachyOS users get the Turkish-locale correction before the next upstream release lands here. Once the relevant upstream release is packaged, we can drop this patch.

PR #1681, opened 2026-08-06
mesa: update to v26.2.0

witchlliee:


ptr1337:
You need to update the rust crates too

witchlliee:
> You need to update the rust crates too

~~I think I will close this one because I still have no idea in how to check for which ones needs updating and to what version. Will have to figure that out later.~~

ptr1337 (approved):

PR #1697, opened 2026-08-09
nvidia-utils: Do not force nvidia-uvm module load

ventureoo:
See https://gitlab.archlinux.org/archlinux/packaging/packages/nvidia-utils/-/merge_requests/33 for details.


nvidia/nvidia-utils/PKGBUILD
vnepogodin: we should remove that one then? if we want to softload it based on dep instead of force loaded

vnepogodin (changes requested):

nvidia/nvidia-utils/PKGBUILD
vnepogodin: +should rebase one arch one at all

ventureoo: > we should remove that one then? if we want to softload it based on dep instead of force loaded

Somehow didn't see this while moving changes from Arch's MR. Done.

vnepogodin (approved):
LGTM

PR #1748, opened 2026-08-22
zlib-ng: add versioned provides

### LINKED ISSUE #1747: [Bug]: zlib-ng-compat cannot satisfy versioned zlib dependencies
### Package name

zlib-ng-compat

### Package version

2.3.3-2

### Issue type

Packaging error (wrong deps, missing files, etc.)

### Description

`zlib-ng-compat` is intended to replace `zlib` and currently sets:

```bash
provides=('zlib' 'libz.so')
conflicts=('zlib')
replaces=('zlib')
```

The `zlib` provision is unversioned. This leads pacman to not consider it sufficient for a dependency such as `zlib>=1.2.3`.

`replaces=('zlib')` does not change this. `replaces` handles package replacement, but dependency resolution uses `provides`, and an unversioned provision cannot satisfy a versioned dependency.

The compat build reports the zlib compatibility version as `1.3.1.zlib-ng` through both `ZLIB_VERSION` and pkg-config, so I think this should be:

```bash
provides=('zlib=1.3.1.zlib-ng' 'libz.so')
```

The same issue exists in `lib32-zlib-ng-compat`, which should provide the corresponding version for `lib32-zlib`.

This causes a hard failure on CachyOS because `lib32-zlib-ng-compat` explicitly depends on `zlib-ng-compat`. If another package requires a versioned `zlib`, pacman tries to install regular `zlib`. That conflicts with `zlib-ng-compat`, but removing `zlib-ng-compat` then breaks the dependency from `lib32-zlib-ng-compat`.

### Steps to reproduce

1. Install `zlib-ng-compat` and `lib32-zlib-ng-compat`.

2. Run:

   `pacman -T 'zlib>=1.2.3'`

3. Pacman prints `zlib>=1.2.3`, so it considers the dependency unsatisfied even though `zlib-ng-compat` is installed.

4. In a transaction containing a package which requires `zlib>=1.2.3`, pacman therefore tries to install regular `zlib`. This conflicts with `zlib-ng-compat`, and removing `zlib-ng-compat` then breaks `lib32-zlib-ng-compat`.

### Relevant logs or error output

```shell
$ pacman -T 'zlib>=1.2.3'
zlib>=1.2.3

:: zlib-1:1.3.2-3 and zlib-ng-compat-2.3.3-2 are in conflict. Remove zlib-ng-compat? [y/N] y
error: failed to prepare transaction (could not satisfy dependencies)
:: removing zlib-ng-compat breaks dependency 'zlib-ng-compat' required by lib32-zlib-ng-compat
```

### CachyOS bug report

N/A - this is a package metadata issue and can be reproduced directly with pacman -T. It is not hardware or kernel dependent.

### Additional context

I reported the underlying unversioned `provides=('zlib')` to Arch upstream:

https://gitlab.archlinux.org/archlinux/packaging/packages/zlib-ng/-/work_items/8

The report was closed and upstream does not seem interested in adding a versioned `provides`, so this probably needs to be carried downstream by CachyOS.

The AUR package mentioned in that report was only a concrete example which exposed the problem. The issue itself is independent of AUR and can be reproduced directly with `pacman -T`.

I also checked the current Arch `core`, `extra` and `multilib` databases. None of the official packages currently require a versioned `zlib`, so official Arch packages do not trigger this today. Third-party packages can, and on CachyOS the explicit `lib32-zlib-ng-compat -> zlib-ng-compat` dependency turns pacman's fallback to regular `zlib` into an unresolvable transaction.

RubenKelevra:
Fixes #1747

### DIFF
--- a/zlib-ng/lib32-zlib-ng/.SRCINFO
+++ b/zlib-ng/lib32-zlib-ng/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = lib32-zlib-ng
 	pkgdesc = zlib replacement with optimizations for next generation systems - 32-bit
 	pkgver = 2.3.3
-	pkgrel = 2
+	pkgrel = 3
 	url = https://github.com/zlib-ng/zlib-ng
 	arch = x86_64
 	license = custom:zlib
@@ -22,7 +22,7 @@ pkgname = lib32-zlib-ng-compat
 	pkgdesc = zlib replacement with optimizations for next generation systems - 32-bit (zlib compat)
 	depends = zlib-ng-compat
 	depends = lib32-zlib-ng
-	provides = lib32-zlib
+	provides = lib32-zlib=1.3.1
 	provides = libz.so
 	conflicts = lib32-zlib
 	replaces = lib32-zlib
--- a/zlib-ng/lib32-zlib-ng/PKGBUILD
+++ b/zlib-ng/lib32-zlib-ng/PKGBUILD
@@ -8,7 +8,8 @@ pkgname=(
   lib32-zlib-ng-compat
 )
 pkgver=2.3.3
-pkgrel=2
+pkgrel=3
+_zlibver=1.3.1
 pkgdesc='zlib replacement with optimizations for next generation systems - 32-bit'
 url='https://github.com/zlib-ng/zlib-ng'
 arch=('x86_64')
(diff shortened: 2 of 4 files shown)


ptr1337 (approved):

1Naim (approved):

simiscoool-afk:
why is this not merged yet?

RubenKelevra:
> why is this not merged yet?

@ptr1337 / @1Naim can you have another look?

PR #1770, opened 2026-08-25
rare-bin: add package

loathingKernel:


ptr1337:
Why?

loathingKernel:
> Why?

because it doesn't need `python313`, it is already included with the artifact from my releases. I made this before noticing the other PR was merged. We can ignore this for now

ptr1337:
> > Why?
> 
> because it doesn't need `python313`, it is already included with the artifact from my releases. I made this before noticing the other PR was merged. We can ignore this for now

Theyre both already in the repository, dont think its really needed.

PR #1795, opened 2026-09-01
cachyos-settings: Enable oomd by default

ventureoo:
Since we have removed le9uo patchset from our kernel there are many similiar reports about system freezing due to out of memory sitations. Installing proper OOM killer should eliminate such issues.


ptr1337:
> Since we have removed le9uo patchset from our kernel there are many similiar reports about system freezing due to out of memory sitations. Installing proper OOM killer should eliminate such issues.

Why specifically earlyoom? Whats wrong with systemd-oomd?

ventureoo:
> > Since we have removed le9uo patchset from our kernel there are many similiar reports about system freezing due to out of memory sitations. Installing proper OOM killer should eliminate such issues.
> 
> Why specifically earlyoom? Whats wrong with systemd-oomd?

Nothing actually wrong with systemd-oomd, it's just systemd-oomd requires some addtional setup for the system:

>      Be aware that if you intend to enable monitoring and actions on user.slice, user-$UID.slice, or their  ancestor  cgroups,  it  is
>      highly recommended that your programs be managed by the systemd user manager to prevent running too many processes under the same
>      session scope (and thus avoid a situation where memory intensive tasks trigger systemd-oomd to kill everything under the cgroup).
>      If you're using a desktop environment like GNOME or KDE, it already spawns many session components with the systemd user manager.

Most of desktop environments unlike GNOME and KDE are not fully integrated with systemd and don't spawn user apps within seperate slice, so it won't work properly and systemd-oomd often kills entire session instead of exact greedy process.

ventureoo:
v2: earlyoom replaced with systemd-oomd, but it depends on new version with https://github.com/CachyOS/CachyOS-Settings/pull/269

ptr1337 (approved):

PR #1819, opened 2026-09-05
nvidia-470xx-utils: 470.256.02-21, add kernel 7.3 support 

simiscoool-afk:
added the 7.3 patch


ventureoo:
Why it was marked as draft?

simiscoool-afk:
i was unsure about why `disable-objtool-override.patch` was not added before but due to the age of this driver and it's hardware it's better to switch to nouveau anyways.

PR #1823, opened 2026-09-08
Update dms-shell

phoropter:


esanchezm:
@phoropter 1.6.1 just released https://github.com/AvengeMedia/DankMaterialShell/releases/tag/v1.6.1

phoropter:
> @phoropter 1.6.1 just released https://github.com/AvengeMedia/DankMaterialShell/releases/tag/v1.6.1

Thanks, fixed

ViniciusIth:
[Instructions from 1.6.0](https://github.com/AvengeMedia/DankMaterialShell/releases/tag/v1.6.0)
> Package Maintainers - PLEASE READ
DMS 1.6 changes the way DMS is packaged, this changes the build process slightly.

> Breaking, Requires Change:

> Quickshell UI is embedded in the binary and no longer needs to be installed to to /usr/share/quickshell/dms . Remove this step from your build script
Add -tags withshell to the normal build tags. Run make sync-shell prior to running go build from the core directory
Checkout must also checkout the dank-qml-common submodule before building. git submodule update --init --recursive
dank-greeter is no longer a part of DMS, it is a standalone app: https://github.com/AvengeMedia/dank-greeter

> Informational:

> dgop, curl are no longer hard dependencies. They can be removed from DMS dependency list
All of the Dank Desktop packages are released together from this point forwards. For example, DMS v1.6.0 was released which also means dgop v1.6.0, dsearch v1.6.0, dcal v1.6.0, and dgreeter v1.6.0 were released alongisde it.

There are more steps for updating to 1.6+ than just updating the tag.

phoropter:
Thanks @ViniciusIth, it should be good now.

ViniciusIth:
You're welcome.

Unrelated, but was the entire change vibe coded? I noticed several updates to `.SRCFILE` without updates to `PKGBUILD` when it is generally generated by running `makepkg --printsrcinfo > .SRCINFO`. I also noticed that both sums are marked as skipped, which is strange considering that `updpkgsums` would've checked the first source (because it is bound by the tag), something that AI agents generally forget to do.

dms-shell/PKGBUILD
```diff
@@ -3,41 +3,50 @@
...
-            'systemd: Required for network managementa via systemd-networkd'
+            'systemd: Required for network management via systemd-networkd'
             'wtype: Paste from clipboard history and plugins')
-makedepends=(go)
-_archive="$_pkg1-$pkgver"
-source=("$url/archive/v$pkgver/$_archive.tar.gz")
-sha256sums=('2d665870f7019700fb38d59d043202e298e1aad1d3f1cb96360a75acbd3be9ca')
+makedepends=(git
+			 go)
+source=("$_pkg1::git+$url.git#tag=v$pkgver"
```
witchlliee: Why are you fetching it from git/tag now?

witchlliee:
Please don't AI generate PKGBUILD updates. Updating them is a simple as `pkgctl version upgrade` while in the PKGBUILD dir.

phoropter:
No, I fumbled through this manually trying to reference how to do things and fixing when I figured something else out. I did it all in GitHub editor. Happy to take direction, I'm obviously not experienced in this.

witchlliee:
> No, I fumbled through this manually trying to reference how to do things and fixing when I figured something else out. I did it all in GitHub editor. Happy to take direction, I'm obviously not experienced in this.

Ah, okay. It's very simple tho:

Install the tools:
```
sudo pacman -S devtools
```

And while in the dms-shell dir:
```
pkgctl version upgrade
```

Regen the .SRCINFO:
```
rm .SRCINFO
makepkg --printsrcinfo >> .SRCINFO
```

phoropter:
Thanks, that looked to be pretty much the same as what I had, I still brought it in though.

phoropter:
> I also noticed that both sums are marked as skipped

I wasn't sure on this as it's pulling from source. The different -git pkgbuilds I looked at had SKIP for the sha256sum.

ViniciusIth:
> I wasn't sure on this as it's pulling from source. The different -git pkgbuilds I looked at had SKIP for the sha256sum.

This is because they are rolling builds, you can't know the sum ahead of time because they change with the commits. Your build pins the tag with `#tag=v$pkgver`, so it is possible to calculate. Notice that the commons doesn't pin (which I am thinking if it could be a problem in the future), hence skip makes sense here.

Sorry if it seemed like I was pointing at you, just wanted to make sure you at least researched what you were doing, and because you talked about other -git packages, now I do believe you did.

1Naim (commented):
I sugggest syncing with Arch Linux.

ViniciusIth:
I really don't like 'arch/extra' approach, using sed to modify like they do

```shell
	sed -i \
		-e '/@update-desktop-database/d' \
		-e '/@gtk-update-icon-cache/d' \
		-e '/^install:/s/install-shell //' \
		Makefile
```

seems a good way to have trouble in the future. I also don't understand the point of it, considering it is later built manually with Golang instead of using the makefile.

But it is up to you guys now. Whatever updates faster, cause I would love to sync back my computer to cachyos packages. If @phoropter ends up being accepted, I think it would be better to squash the changes.

ptr1337 (approved):

PR #1824, opened 2026-09-08
cachyos-drm-awaiter: emit target on NVIDIA-only systems

ahzaidi:
On a machine where the only GPU is NVIDIA, the generator never writes `drm-module-load.target`. It stores the NVIDIA module in `nvidia_module` (so it can be ordered after an iGPU driver) and only appends it to `modules` after the `[ -z "${modules[0]}" ]` early return. With no AMD or Intel GPU that array is empty, so the function returns before the append.

This patch moves the NVIDIA merge above the early return. The postpone loop runs before the append now, so the `!= nvidia_module` guard goes away. Hybrid and AMD/Intel-only machines produce the same output as before.

How I hit it: installed 1-1 on my desktop (RTX 5070 Ti, proprietary driver, no iGPU), rebooted, and `/run/systemd/generator/` had no target and no `.wants` dirs. `sys-module-nvidia_drm.device` was there, so the udev half works. Running the script by hand with `bash -x` shows `nvidia_module=nvidia_drm` followed straight by `return 0`. With the patch installed and a real reboot, `/run/systemd/generator/drm-module-load.target` is there with `After=sys-module-nvidia_drm.device` and both symlinks, the target went active at 4.2s and plasmalogin started 3ms after it. No change in boot time since the module is still in the initramfs here.

Also bumped pkgrel and regenerated .SRCINFO. The old .SRCINFO had a different checksum for the udev rule than the PKGBUILD did, so that gets corrected as a side effect.

Two things I noticed but left alone, since they are separate bugs: `is_module_loadable nouveau` can never match because nouveau's aliases use `d*` wildcards and the check greps for the exact device id, so the nouveau branch is dead; and `generator_dir` is not validated, so running the script with no argument writes into `/`. Happy to send those separately if you want.

I used an LLM to help find and draft this. The repro and the test above are from my own machine.

### DIFF
--- a/cachyos-drm-awaiter/.SRCINFO
+++ b/cachyos-drm-awaiter/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = cachyos-drm-awaiter
 	pkgdesc = Configuration for proper systemd chain-loading of DRM modules
 	pkgver = 1
-	pkgrel = 1
+	pkgrel = 2
 	url = https://github.com/CachyOS/CachyOS-PKGBUILDS/tree/master/cachyos-drm-awaiter
 	arch = any
 	license = GPL-3.0-or-later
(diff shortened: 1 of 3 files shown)


ventureoo (approved):
Oops, yes, that certainly make sense. Thank you for pointing out!

ventureoo:
> Two things I noticed but left alone, since they are separate bugs: is_module_loadable nouveau can never match because nouveau's aliases use d* wildcards and the check greps for the exact device id, so the nouveau branch is dead; and generator_dir is not validated, so running the script with no argument writes into /. Happy to send those separately if you want.

The script should be run by systemd-only, so I don't think we really need to validate it. As for nouveau stuff, yeah, it will be appreciated, since modalias stuff is a bit fragile.

ventureoo:
> Two things I noticed but left alone, since they are separate bugs: is_module_loadable nouveau can never match because nouveau's aliases use d* wildcards and the check greps for the exact device id, so the nouveau branch is dead;

Can you please check if https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/1862 actually helps with this?

PR #1827, opened 2026-09-09
archlinux-keyring: update to v20260909

witchlliee:


ptr1337:
Sorry, i did not see it

PR #1846, opened 2026-09-11
linux-firmware: Update to v20260910

witchlliee:


ptr1337:
Did it manually, since the PR was not fully done and missing changes from Arch and removals in prepare section

PR #1848, opened 2026-09-11
cachyos-gaming-applications: Remove lutris in favor of faugus-launcher

witchlliee:
Lutris doesn't hold well against faugus-launcher. Faugus already does a way better job and fills the gap which once lutris did. I don't think it has a space in the cachyos-gaming-applications anymore.

### DIFF
--- a/cachyos-gaming-applications/.SRCINFO
+++ b/cachyos-gaming-applications/.SRCINFO
@@ -9,7 +9,6 @@ pkgbase = cachyos-gaming-applications
 	depends = heroic-games-launcher
 	depends = faugus-launcher
 	depends = lib32-mangohud
-	depends = lutris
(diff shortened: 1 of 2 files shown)


NextWork123:
I don't believe Faugus serves as a full replacement for Lutris. Lutris provides critical functionality that Faugus currently lacks, notably custom installation scripts, deeper configuration options, and a more intuitive UI. The best approach would be to retain both and allow users to choose based on their workflow.

but i will leave ptr and others decide.

witchlliee:
It does have custom installation scripts for launchers (Battle.net, Epic, GOG, etc) and it does have a handful of user configurations for environment variables, arguments, tweaks for proton, etc. It provides more proton options as well for user's choice. 

image

PR #1849, opened 2026-09-11
Remove davinci-resolve

simiscoool-afk:
Not allowed to redistribute.


ptr1337:
I need the PKGBUILD to be able to pull the ZIP. Ill start maintaining it soon again

